PR #25, opened 2018-06-20
Create README.md

clausdenk:
Maybe it might be valuable to have the "hydra serve --help" output in some appendix or wherever in the docs. This could be referenced at different places in the docs.


arekkas:
Thank you!

PR #26, opened 2018-06-25
FIx misspell api doc

ionutzp:


arekkas:
Thanks! Unfortunately this is the wrong place to fix it, the proper one is [here](https://github.com/ory/hydra/blob/master/consent/handler.go#L181).

PR #42, opened 2018-08-10
Oathkeeper: Fix typo in rules documentation

vHanda:
It doesn't make sense to say "if b can handle this, then 'a' and 'b' will be ignored".


arekkas:
Thanks!

PR #50, opened 2018-09-06
add --token-endpoint-auth-method if public clients

HIRANO-Satoshi:
I was  confused with an error message when I connected from a public client (AppAuth-JS example).

I needed to add `--token-endpoint-auth-method none`.


aeneasr:
Thanks!

PR #71, opened 2018-11-12
fixed example for jwt authenticator

stszap:
* removed unnecessary comma
* fixed field name for aud check
* fixed values for aud and iss checks


aeneasr:
Nice, thanks!

PR #88, opened 2018-12-13
Replace /jwks endpoint with /keys

condemil:
Accordring to REST API doc https://www.ory.sh/docs/hydra/sdk/api#retrieve-a-json-web-key-set


aeneasr:
Thank you!

PR #94, opened 2018-12-17
Fixed documentation error, updated correct response from the API

hassanfarid:
I found this bug while reading through the documentation. Thought I can fix it quickly.


aeneasr:
Thank you!

PR #98, opened 2019-01-09
Fixed configure-deploy document error & typo

DavidSonoda:
I went through the examples given in the configure-deploy chapter and found an environment variable name error causing the `HYDRA_ADMIN_URL` cannot be correctly set in login-consent app, resulting a null property error in the consent app when testing the OAuth2 flow out.


aeneasr:
Thank you!

PR #110, opened 2019-03-13
5min-tutorial: update docker exec instructions

lopezator:
Replace over-complicated and error prone `docker ps -f name=hydra_hydra_1 -q` exec intructions (that could also cause name collisions) for integrated docker-compose exec functionality: `docker-compose exec hydra`.


aeneasr:
This is gold, thank you!

PR #111, opened 2019-03-14
Acceptlogin required subject

kminehart:
I didn't see any scripts to generate `docs/hydra/sdk/` from this JSON file, so I just assumed it was created manually.


aeneasr:
Thank you! However, this stuff is being auto-generated via the CI and the corresponding swagger definitions in the actual code base. To change the field, go [here](https://github.com/ory/hydra/blob/master/consent/types.go#L152) and add `required: true` - everything else will be done by magic (automation) in the next release. See an example of this here: https://goswagger.io/use/spec/model.html#example

PR #116, opened 2019-04-03
Fix options for StringMatchCondition

minchao:
Ref https://github.com/ory/keto/issues/82#issuecomment-479322618


aeneasr:
Thank you!

PR #132, opened 2019-04-25
Update README.md: Add caveats for static resources

ashutoshgngwr:
This should be helpful in warning future editors against the pitfall I just encountered. :baby:


aeneasr:
Thank you!

PR #136, opened 2019-04-29
Update linux install path

aeneasr:
Signed-off-by: aeneasr <aeneas@ory.sh>


aeneasr:
Fixed by another PR without conflicts

PR #139, opened 2019-05-01
ttl is a top-level config field

MDrollette:
don't nest ttl under oauth2 section of config.yaml

This was edited directly on github.com, seems there's some additional line-endings being changed here as well 🤷‍♂


aeneasr:
Thank you! This file is actually being synched from [here](https://github.com/ory/hydra/blob/master/docs/config.yaml) so this needs to be updated there or the changes will be lost :)

PR #142, opened 2019-05-11
oauth2.md: fix logout docs

mkontani:
Thanks for helpful docs. I found a few typos.

logout flow diff: 
(from) Logout Provider->>ORY Hydra: Redirects end user to redirect url with login verifier
(to) Logout Provider->>ORY Hydra: Redirects end user to redirect url with logout challenge


aeneasr:
Thank you!

PR #149, opened 2019-05-20
Fix Oathkeeper proxy labels

kevgo:
Fixes wrong labels at https://www.ory.sh/docs/oathkeeper/#reverse-proxy

image


kevgo:
FYI the image changes are easiest to review when switching the image diff to the "rich diff", then below the image preview to "Onion skin", then drag the slider back and forth.

aeneasr:
Ah yeah, awesome - thank you!

PR #150, opened 2019-05-20
Simple grammar improvements for Oathkeeper intro

kevgo:
Mostly replaces passive voice (which is a no-no in English) with active voice. I also added mention of attribute-based access control as well as usage of some of its terminology (policy enforcement point, policy decision point). I found this model to be one of the more intuitive ones. Feel free to disagree on this or anything else, then I'll take these parts out.

### DIFF
--- a/docs/oathkeeper/index.md
+++ b/docs/oathkeeper/index.md
@@ -50,7 +50,7 @@ ORY Oathkeeper does not forward the request and instead returns an error
 message.
 
 <a href="https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgQyBhcyBDbGllbnRcbiAgICBwYXJ0aWNpcGFudCBPIGFzIE9hdGhrZWVwZXIgUHJveHlcbiAgICBwYXJ0aWNpcGFudCBBIGFzIFByb3RlY3RlZCBTZXJ2ZXIvQVBJXG4gICAgQy0-Pk86IEhUVFAgUmVxdWVzdFxuICAgIE8tLT4-TzogQ2hlY2sgaWYgcmVxdWVzdCBpcyBhbGxvd2VkXG4gICAgYWx0IGlzIG5vdCBhbGxvd2VkXG4gICAgTy0-PkM6IFJldHVybiBIVFRQIEVycm9yIFxuICAgIGVsc2UgaXMgYWxsb3dlZFxuICAgIE8tPj5BOiBGb3J3YXJkIEhUVFAgUmVxdWVzdCBcbiAgICBBLT4-TzogUmV0dXJuIEhUVFAgUmVzcG9uc2VcbiAgICBPLT4-QzogUmV0dXJuIEhUVFAgUmVzcG9uc2VcbiAgICBlbmQiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ">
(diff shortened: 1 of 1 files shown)


docs/oathkeeper/index.md
```diff
@@ -3,12 +3,15 @@ id: index
...
-
-The implemented problem domain and scope is called Zero-Trust Network
+ORY Oathkeeper authorizes incoming HTTP requests.
+It can be the Policy Enforcement Point in your cloud architecture, 
+i.e. a reverse proxy in front of your upstream API or web server 
+that forwards authorized requests and rejects unauthorized ones.
+If you want to use another API Gateway (Kong, Nginx, Envoy, AWS API Gateway, ...),
+Oathkeeper can also plug into that and act as its Policy Decision Point.
+
+The implemented problem domain and scope is part of [attribute-based access control](https://en.wikipedia.org/wiki/Attribute-based_access_control),
```
aeneasr: Oathkeeper is agnostic to access control mechanisms (ABAC, RBAC, ACL, ...) but instead uses the concept of enforcing policies/permissions which can be implemented in a variety of ways. Currently supported are "allow all requests", "deny all requests", and ORY Keto as a Policy Decision Point.

docs/oathkeeper/index.md
```diff
@@ -100,10 +103,10 @@ request to ORY Oathkeeper.
 
 ### Access Control Decision API
 
-The Access Control Decision API is exposed at `/decision` of the ORY Oathkeeper
-API port. It follows best-practices and works with most (if not all) modern API
-Gateways and Reverse Proxies. The endpoint matches every sub-path and HTTP
-Method:
+The ORY Oathkeeper Access Control Decision API follows best-practices and works with most (if not all)
```
aeneasr: Is it clear enough that `/decision` is exposed at the API port, not the reverse proxy port?

docs/oathkeeper/index.md
```diff
@@ -100,10 +103,10 @@ request to ORY Oathkeeper.
...
-The Access Control Decision API is exposed at `/decision` of the ORY Oathkeeper
-API port. It follows best-practices and works with most (if not all) modern API
-Gateways and Reverse Proxies. The endpoint matches every sub-path and HTTP
-Method:
+The ORY Oathkeeper Access Control Decision API follows best-practices and works with most (if not all) 
+modern API gateways and reverse proxies.
+To verify a request, send it to the `decision` endpoint.
+It matches every sub-path and HTTP Method:
 
 - `GET /decision/v1/api`
```
aeneasr: I actually noticed that the endpoint is `/decisions`, not `/decision`. While you're at it would you mind updating this?

docs/oathkeeper/index.md
kevgo: Okay, I have reverted this change. Thanks for the background!

docs/oathkeeper/index.md
kevgo: Added wording to make that clear.

docs/oathkeeper/index.md
kevgo: done

kevgo:
All comments addressed. Thanks for the tip for formatting the code! I was editing this directly using the GitHub UI. What do you think about auto-formatting branches using something like [Prettifier](https://github.com/kevgo/prettifier)?

aeneasr:
Thank you!

 > I was editing this directly using the GitHub UI. What do you think about auto-formatting branches using something like Prettifier?

Can that work with forks?

kevgo:
Good point. I guess people who fork this have to install the bot on their fork. It might still be nice to tell them how to format their contributions on the CLI, and possibly point them to install the bot if they are lazy or mostly use the UI (like me in both cases). I'll send a PR.

PR #160, opened 2019-05-22
Extract Prettier config into prettierrc

kevgo:
I made it in JSON because I know you don't like YML  :)

It somehow also re-formatted a bunch of MD files, which looks legit.
No content changes have been made except Prettier changing double-quotes to single-quotes in the code blocks inside Markdown.

I have removed the `printWidth` option because the default is already 80 characters.


kevgo:
Hmm, our friend `prettifier[bot]` cannot sign this since he has no hands... I own it. Can I sign instead? Or can you whitelist it?

aeneasr:
That's pretty cool (the bot!). Does that exist for go etc too? Also, do I need to install it for this repo separately?

kevgo:
Cool, I'm glad you like it! Right now it's a prototype that only runs Prettier. But it's certainly possible to add support for other languages, that's a great idea! I'll have to see when I get to it since currently I'm pretty much traveling the entire time.

kevgo:
Yes, you have to give Prettifier access to your repos. Either for all repos in your Github org, or just a list of them.

aeneasr:
Oh I didn't see that you wrote that! Very cool, 10/10

PR #161, opened 2019-05-22
Prettier JSON files

kevgo:
Prettier also allows formatting of JSON files.


website/package.json
```diff
@@ -15,6 +15,7 @@
     "serve": "serve -p 9009 ./build",
     "format:markdown": "prettier --write --print-width 80 --prose-wrap always \"{../docs/**,..}/*.md\"",
     "format:code": "prettier --single-quote --trailing-comma es5 --write \"./**/*.{js,css,html}\"",
+    "format:json": "prettier --write '../**/*.json' '!../_legacy/**'",
```
kevgo: I don't format the `_legacy` directory because I guess it's old and we want to keep this fast since it runs in a pre-commit hook.

aeneasr: :)

PR #163, opened 2019-05-27
Delete extra link for non Refresh Token issuing examples.

adityasrini:
This appears to be a typo. Either the extra link was not needed or another link should have been in its place. Open to edits from maintainers.


aeneasr:
Thank you!

PR #165, opened 2019-06-10
Fixes for parts of Keto documentation

kevgo:
I tried to keep it to simple corrections without changing too much. Sometimes I saw an opportunity to move content around in order to say things in a more intuitive order or avoid saying the same thing twice in different places.

### DIFF
--- a/docs/keto/configuration.md
+++ b/docs/keto/configuration.md
@@ -10,7 +10,7 @@ You can configure ORY Keto via a config file or environment variables.
 ## Environment Variables
 
 You can provide all configuration settings using environment variables.
-Environment variables always override values from the configuration file. 
+Environment variables always override values from the configuration file.
 
 Set an environment variable on Linux / OSX:
 
@@ -45,11 +45,12 @@ Calling Docker with environment variables set:
 $ docker run -e MY_ENV_VAR=foo oryd/keto:...
 ```
 
-To provide nested configuration settings as environment variables, format them in screaming snake case:
-`serve.public.port.something_else` becomes `SERVE_PUBLIC_PORT_SOMETHING_ELSE`.
+To provide nested configuration settings as environment variables, format them
+in screaming snake case: `serve.public.port.something_else` becomes
+`SERVE_PUBLIC_PORT_SOMETHING_ELSE`.
 
-To provide multiple values for a key, separate them with a comma.
-Assuming we have this configuration setting:
+To provide multiple values for a key, separate them with a comma. Assuming we
+have this configuration setting:
 
 ```yml
 secrets:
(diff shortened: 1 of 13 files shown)


docs/keto/configuration.md
```diff
@@ -3,82 +3,92 @@ id: configuration
 title: Configuration
 ---
 
-```yaml
 # ORY Keto Configuration
-#
-#
+
+You can configure ORY Keto via a config file or environment variables.
```
kevgo: This was all in comments in the config file. I have extracted the parts that don't seem related to configuring Keto, like how to set up env variables or Docker, into Markdown documentation on top of this file.

docs/keto/engines/acl.md
```diff
@@ -9,16 +9,16 @@ is a matrix of users and permissions:
 
 |       | blog_post.create | blog_post.delete | blog_post.modify | blog_post.read |
 | ----- | ---------------- | ---------------- | ---------------- | -------------- |
-| alice | yes              | yes              | yes              | yes            |
-| bob   | no               | no               | no               | yes            |
-| peter | yes              | no               | yes              | yes            |
+| Alice | yes              | yes              | yes              | yes            |
```
kevgo: One liberty I took is capitalizing people's names, at least in text. In code, they could still be lowercase since it's data. But in Markdown it reads really weird if they are lowercase.

docs/keto/engines/acp-ory.md
```diff
@@ -547,25 +526,23 @@ developed over the years at ORY.
 
 ### URNs
 
-> “There are only two hard things in Computer Science: cache invalidation and
-> naming things.” -- Phil Karlton
+> “There are only two hard things in Computer Science: cache invalidation,
+> naming things, and off-by-one errors.” -- Phil Karlton
```
kevgo: Added this joke, not sure if it is from Phil Karlton. But I think it adds a little breath of fresh air while reading through this long piece of security documentation.

docs/keto/engines/index.md
```diff
@@ -1,24 +1,18 @@
 ---
 id: index
-title: Introduction
+title: Access Controle Engines - Introduction
```
kevgo: This might be too long to render properly. But just "introduction" seems not enough info here.

kevgo:
Sorry for the long PR. This has accumulated over several plane flights without Wifi. I'll send additional changes in a separate branch.

docs/keto/configuration.md
aeneasr: This is being auto-generated from [here](https://github.com/ory/keto/blob/master/docs/config.yaml) and thus overwritten on every git push to master. The script doing that is this one: https://github.com/ory/keto/blob/master/scripts/run-configuration.sh

Not a merge blocker, but changes will be lost on next push to keto.

docs/keto/engines/acp-ory.md
aeneasr: :)

PR #172, opened 2019-07-04
Update link to renamed docker-compose file

shankardevy:
Looks like the docker-compose.yml is renamed and split into multiple files. Linking to the appropriate file.


aeneasr:
Thanks!

PR #173, opened 2019-07-05
Add 409 to accept {consent|login} responses

cburmeister:
This commit points out that an API consumer can receive a 409 when
accepting either the consent or login requests.


aeneasr:
Hi there! Thank you for the change! The Swagger docs are being autogenerated so the right place to add the 409 error would be

* https://github.com/ory/hydra/blob/master/consent/handler.go#L304
* https://github.com/ory/hydra/blob/master/consent/handler.go#L390
* https://github.com/ory/hydra/blob/master/consent/handler.go#L528
* https://github.com/ory/hydra/blob/master/consent/handler.go#L606
* https://github.com/ory/hydra/blob/master/consent/handler.go#L669
* https://github.com/ory/hydra/blob/master/consent/handler.go#L704

PR #174, opened 2019-07-08
Update api.md

fatalc:
correct a table style


aeneasr:
Thank you for your contribution, unfortunately the part modified is auto-generated. In this case it seems to be a bug in the generator for this specific table. I will leave this PR open to remind me to investigate why that happens and fix the generator!

aeneasr:
I've resolved the table-generator issue! Thank you

PR #188, opened 2019-08-06
housekeeping: Fix typos

goknsh:
Fixes typos in the following directories:
* `apis/`
* `docs/`


apis/hydra.json
```diff
@@ -105,7 +105,7 @@
         }
       },
       "post": {
```
aeneasr: These changes need to be made here: https://github.com/ory/hydra/blob/master/jwk/handler.go#L145 :)

Otherwise they will be overwritten by the next automatic API generation :)

PR #191, opened 2019-08-11
Use `--help` instead of `help`

jmonteiro:
In the latest version using `help` causes the help message to be prefixed with `Config file not found because "Config File ".hydra" Not Found in "[/]""` on a new install.


aeneasr:
Thank you! The issue should have been resolved on master (pending release) so `help` is still a valid command :) Therefore closing this as this behavior will go away once we hit 1.0.1 :)

PR #205, opened 2019-09-25
Updated config.yaml example

GraafG:
If config not changed to my edit you will get this error:

time="2019-09-25T09:08:51Z" level=fatal msg="The configuration is invalid and could not be loaded." error="must validate at least one schema (anyOf)" validation_error[0]="mutators.header: Must validate at least one schema (anyOf)" validation_error[1]="mutators.header.config: config is required"


aeneasr:
Thank you!

PR #208, opened 2019-10-07
Remove Authentication Note (Token Introspection)

rwenz3l:
The Token Introspection Endpoint is only available to admins
and therefore should not require any authentication.
See: https://github.com/ory/hydra/issues/1520

### DIFF
--- a/docs/hydra/sdk/api.md
+++ b/docs/hydra/sdk/api.md
@@ -8037,10 +8037,6 @@ token was not granted one of the scopes, the result of active will be false.
 }
 ```
 
-<aside class="warning">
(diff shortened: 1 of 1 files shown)


docs/hydra/sdk/api.md
```diff
@@ -8037,10 +8037,6 @@ token was not granted one of the scopes, the result of active will be false.
 }
 ```
 
-<aside class="warning">
```
aeneasr: This is generated from code [here](https://github.com/ory/hydra/blob/master/oauth2/handler.go#L390-L392), so it needs to be fixed there.

PR #214, opened 2019-10-14
Remove unnecessary paragraph in Hydra API docs

slashmo:
The removed paragraph makes sense in the `/health/alive` and `health/ready` endpoints documentation but feels misplaced for documenting the `/version` endpoint.


aeneasr:
Thank you ! However, as noted at the top of the document, this file is being auto-generated. The appropriate line to fix this is here: https://github.com/ory/hydra/blob/master/health/doc.go#L106-L107

PR #216, opened 2019-10-19
removed the token introspection command

bayer:
`hydra clients create` doesnt display the token anymore, so i removed the section dealing with the token

### DIFF
--- a/docs/hydra/5min-tutorial.md
+++ b/docs/hydra/5min-tutorial.md
@@ -84,9 +84,13 @@ $ docker-compose -f quickstart.yml exec hydra \
     --id my-client \
     --secret secret \
     -g client_credentials
+```
 
+If you get an error message about a config file not being found, you can ignore it for now:
+
+```
+Config file not found because "Config File ".hydra" Not Found in "[/]""
 OAuth2 client my-client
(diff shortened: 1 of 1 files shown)


docs/hydra/5min-tutorial.md
```diff
@@ -98,30 +102,6 @@ $ docker-compose -f quickstart.yml exec hydra \
     --client-id my-client \
     --client-secret secret
 
-UDYMha9TwsMBejEvKfnDOXkhgkLsnmUNYVQDklT5bD8.ZNpuNRC85erbIYDjPqhMwTinlvQmNTk_UvttcLQxFJY
```
aeneasr: Please revert this change. The information shown here is correct and important.

bayer: I dont't think the information is correct, since hydra does not display the token after the call - that's why I removed the token inspection command.

aeneasr: The information is 100% correct, if it didn't show the token it was an error (such as misspelling, waiting too long before using the token, ...).

aeneasr:
Thank you!

PR #219, opened 2019-11-03
Clarify Token Instrospection endpoint

florv:
Clarify the situation of the Token Introspection endpoint requiring authentication but having no access-control.

[Related forum discussion](https://community.ory.sh/t/token-introspection-not-working/1330)


aeneasr:
awesome, thanks!

PR #257, opened 2020-01-25
Tc thurs rev2

tacurran:


docs/kratos/concepts/index.md
```diff
@@ -38,170 +38,128 @@ by combining ORY Kratos with [ORY Hydra](http://github.com/ory/hydra) .
...
 and tests critical network flows, system architectures, user flows, protocols, and business logic.
 
-#Begin edits @tc
-
 ### Full-stack Identity and Access Management (IAM)
 
-*Disclaimer: There are no product and projects names from this section because we do not believe that it is
-the right place. However, all the problems we describe have been observed in real products and projects. If you're
-interested how ORY Kratos compares to other projects and products, head over to [Comparison](../further-reading/comparison.md)*
+*Disclaimer: There are neither product nor projects names in this section. This section's goal is to describe circumstances and problems that mostly stem from the community's experience. For information purposes, there is a preliminary comparision of ORY Kratos and other projects and products available at [Comparison](../further-reading/comparison.md)*.
```
aeneasr:
```suggestion
*Disclaimer: There are neither product nor project names in this section. This section's goal is to describe circumstances and problems that mostly stem from the community's experience. For information purposes, there is a preliminary comparision of ORY Kratos and other projects and products available at [Comparison](../further-reading/comparison.md)*.
```

aeneasr: Do you think we should keep this? We can also move "Comparison" to the blog but I think a lot of devs actually want to see a technical comparison of the different projects and how they relate.

PR #264, opened 2020-01-30
Update production.md to add a logout endpoint.

ijust:
I try to make a production environment and find that I cannot access a logout endpoint from public internet.
Should I not allow to access this logout endpoint from the internet?


aeneasr:
Thank you!

PR #279, opened 2020-02-26
fix: Markdown table format.

yuan1238y:
Fix chapter "Create an identity" "Responses overview" table format.


aeneasr:
Thank you! This file is actually generated automatically from a swagger definition. But it should definitely not have those lines in there. I will keep this PR open and track it as a bug, thank you!

PR #282, opened 2020-03-06
docs: add global auth session

Sbou:
reference to https://github.com/ory/oathkeeper/pull/358

@aeneasr can you check if it misses something?
Thanks


aeneasr:
You rock!

PR #288, opened 2020-03-27
fix: change an image name

tomoyamachi:
`oryd/hydra-login-consent-node:v1.3.2` does not exist.
https://hub.docker.com/r/oryd/hydra-login-consent-node/tags


aeneasr:
Thank you for bringing this to our attention! I've pushed the image with the correct tag name on docker hub, so the original instructions should work now :)

PR #300, opened 2020-05-21
docs: sqa.md: minor content, markdown tweaks

salewski:
- markdown: one instance of normalizing list syntax (first item in list was
  numbered, but rest of list items were unordered)

- prose: use "SHA-256" everywhere (rather than sometimes saying "sha256")

- prose: change one instance of "ip addresses" to "IP addresses" (for
  consistency with usage elsewhere in the article)

- prose: use "UUID" rather than "uuid" (for consistency -- most acronyms in
  article are in all caps)

- prose: restate: "We built this pipeline the data with the following goals:"
              as: "We built this pipeline with the following goals in mind:"


aeneasr:
Thank you!

PR #349, opened 2020-12-01
docs: fix up contributing

vinckr:
Just some small corrections in the beginning before I put it into the sidebar.


vinckr:
We could also rename this one to "Ory Development Environment" or something similar?
To distinguish it from CONTRIBUTING.md and make clear that this is mostly scripts/tips to build.

PR #367, opened 2021-01-05
Issue 2267 update doc on limit param upper bound

thormarius:
Issue `#2267`: Document that there is an upper bound on `limit` parameter to prevent DoS


aeneasr:
Awesome 🎉 

Thank you for your contribution!

PR #368, opened 2021-01-06
docs: add kratos projects.md

### LINKED ISSUE #366: Include Kratos in projects.md
In [Ecosystem](https://github.com/ory/docs/blob/master/docs/docs/ecosystem/projects.md) an soon-to-be-announced identity management server is mentioned. 
We should update the section to include Kratos with some examples like the other projects. 
I think the text is not bad otherwise, so just a small update would fix it here.

The graphic also needs the updated logos (and maybe the colours of each project?) @jfcurran.

vinckr:
short & sweet?
I didnt want to repeat too much from the Kratos page.
closes #366


docs/docs/ecosystem/projects.md
```diff
@@ -30,10 +30,14 @@ Our products solve the simplest use case and give you the ability to instantly
 ready the system for more complex scenarios without painful and slow upgrade
 processes.
 
-The above-mentioned _identity management server_ is to be announced soon.
+The identity management server **ORY Kratos** enables you to implement login and
+registration in a secure and straightforward way.
```
aeneasr, line 34: Let's add "mfa / 2fa, profile management, session management, user management" to that list

aeneasr (changes requested):

vinckr:
Disregard the mess with the package-lock.json... 

Updated the text and added something about security as well.

PR #382, opened 2021-03-08
docs: add project

vinckr:
Added a community project and removed @k9ert project (for now).

@aeneasr do you have any input on how we could integrate this page into ory/works? 

We could just link it, but imo would be nice to have all that info directly in the ory/works readme too.. 

I could append it, like we do with ADOPTERS.md in the other readmes, what do you think of that approach?

### DIFF
--- a/docs/docs/ecosystem/community.md
+++ b/docs/docs/ecosystem/community.md
@@ -27,9 +27,9 @@ This is a list of those.
 
 - Open source software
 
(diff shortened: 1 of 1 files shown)


aeneasr:
> @aeneasr do you have any input on how we could integrate this page into ory/works?
>
> I could append it, like we do with ADOPTERS.md in the other readmes, what do you think of that approach?

Yeah I think that should work well

PR #414, opened 2021-06-03
docs: registration flow 

vinckr:
Draft PR for the self-service registration flow documentation

https://github.com/ory/docs/commit/ed7cae11a2d7b88a24deb687d28b996589328465 adds the mermaid and code example modules, as well as the pictures and code examples. 
They need to be updated for the SaaS, but I wanted to confirm if it is going in the right direction before doing so.

https://github.com/ory/docs/commit/779cfbdb54ef681da8b9cc73b8fb270db5766b31 adds the doc for registration flow itself. It is basically the OS document, rephrased and changed a bit in some aspects. 

I also left some comments on open questions in the issue [here](https://github.com/ory-corp/product/issues/152#issuecomment-852973618).
I also left a few comments in the registration.mdx

Could you confirm this is going in the right direction @aeneasr ?


aeneasr:
I think this is problematic because we are copy pasting basically stuff from ory/kratos. Next time there's an update everything has to be redone again, or done twice manually which is extremely error-prone. I myself am still finding several broken copy & paste sections in the existing Ory Kratos docs and there the content isn't even 1:1 :)

So I don't know. I think we can merge this as a temporary solution? But really what we need is to figure out a process for taking the open source docs and making them also compatible with the closed source docs.

Maybe we have in the Ory Kratos docs a section which covers both services, then a section which covers e.g. only open source or only cloud.

In the platform docs the links then point to that document.

However, that's probably a bit confusing when browsing the docs for the first time!

Another option is to have, in Ory Kratos, different types of documents and we copy the relevant docs into the platform and remove them before building the docs for open source?

But yeah this needs an automated process because manual copy paste just increases everyone's workload and leads to terrible, outdated, or broken docs!

vinckr:
Most of the content is the same. The biggest change is in the code examples. 

We could copy the text for both and just switch out the code examples for SaaS & OS?
(Is that a problem for SEO?)

I would resolve conflicts / format everything and then tag you for a review ok? @aeneasr 

Then we have content available and can think about a smarter/ less labour intense way to organize it going forward?

aeneasr:
So I think what we need first is a good sidebar / TOC and then we start working on the content. Could you come up with 2-3 proposals? We can keep this PR open but I would put it in "blocked"

aeneasr:
Marking this as  a draft

PR #444, opened 2021-07-26
fix: typo

marneysofia:
"haslimitations" --> "has limitations" added a space between the words.


aeneasr:
Thank you! I just saw your PR after I made a refactoring of the docs structure - the LoC is no longer existent so I will close the PR but thank you very much for the contrib :)

PR #445, opened 2021-07-26
fix: typo

marneysofia:
Changed "This list is going to expand and maintained in the future" to "This list will be expanded and maintained in the future."


aeneasr:
Awesome, thank you! :)

PR #449, opened 2021-07-31
fix:ory-cli-install-use linux title

180909:


zepatrik (approved):
Thanks for fixing the typo, much appreciate your effort :+1:

PR #453, opened 2021-08-04
docs: Custom Identity Model Documentation

vinckr:
closes: https://github.com/ory-corp/cloud/issues/1275

@Benehiko Is the part about "required" correct? I could not get it to work. 
Please let me know, if I should remove it or it is working differently than I thought.

- docs: custom im concept
- docs: custom im guide
- feat: custom im guide screenshots
- docs: add im guide to sidebar


docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,304 @@
...
+  Model.
+
+- Add or remove traits from the Identity Model as desired, more info in the
+  [Customize Fields](#customize-fields) section.
+
+- When finished editing name your custom Identity Model. Input the name in the
+  textbox between the checkbox and the schema.
+
+- Press `Enter` or hit the `Update` button to save it, once you have made your
+  changes. Currently, it is not possible to edit an existing Identity Model, so
```
Benehiko, line 35: Maybe this needs some more details such as

```suggestion
  changes. Currently, it is not possible to edit an existing Identity Model on the same revision. The identity model can still be reused as a template for a following revision.
```

Then maybe have a section explaining what the revision would look like:

Revisions 1:
 - I create a schema with name "MyUsersv1"
 - Create the identity model and save it

Revisions 2:
 - Select the schema "MyUsersv1" as a template
 - Enter a new name (can technically be the same but there will be 2 of them then with different creation times)
 - Make the necessary adjustments to the identity model
 - Hit Save

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,304 @@
...
+
+The
+[`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)
+keyword is used to control the handling of properties whose names are not listed
+in the properties keyword. This usually should be set to false.
+
+```json
+      "additionalProperties": false
+```
+
```
Benehiko, line 212: As far as I can remember this wouldn't make a difference currently to the managed UI - only to custom UI's.

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,304 @@
...
+"properties": {
+        "handle": {
+          "type": "string",
+          "title": "Your GitHub Handle"
+        }
+      }
+    }
+```
+
+or a mobile phone number as `integer`:
```
Benehiko, line 157: phone number as an integer won't be a good user experience.
It will create a input field with a "increase/decrease" button on the side - which looks bad.

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,304 @@
...
+
+```json
+      "additionalProperties": false
+```
+
+## Example Identity Model
+
+The following Identity Model includes first/last and nickname, as well as number
+fields for the phone number and birthday. There are also two true/false fields
+for specifying the newsletter subscription and enterprise status.
```
Benehiko, line 217: integer for phone number isn't ideal

Benehiko (changes requested):
There are some changes that need to be done to the phone number example. We should maybe use something different such as "Age" or decide on a course of action to the UI with the "ingeger" type sine the UX is quite bad on it.

docs/docs/concepts/identity.mdx
```diff
@@ -12,19 +12,22 @@ check out the open source documentation:
...
 ## Available Presets
 
-At the moment we support three basic Identity Model Presets.
+The Ory Cloud provides three basic Identity Model Presets.
 
 ### Only Username
 
-This preset is useful for applications which do not require email addresses and
-a high degree of anonymity.
+This preset is useful for applications that do not require email addresses and a
```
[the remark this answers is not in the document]
vinckr: Yea I was wondering about that as well. Will change the wording 👍

docs/docs/guides/customize-identity-model.mdx
vinckr: Great Idea! I will add such a section.

docs/docs/guides/customize-identity-model.mdx
vinckr: True, I noticed that as well. 
I think `age` would probably a good example for now.

docs/docs/guides/customize-identity-model.mdx
vinckr: I will change it to `age` 🙂 

Do you think it makes sense to include a notice that phone numbers are not really supported right now?

docs/docs/guides/customize-identity-model.mdx
vinckr:
```suggestion
The
[`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties)
keyword is used to control the handling of properties whose names are not listed
in the properties keyword. This currently has no effect and should be set to false.

```json
      "additionalProperties": false
```

Maybe like so? 

I wanted to include it, since not mentioning it at all would also be confusing IMO.

docs/docs/concepts/identity.mdx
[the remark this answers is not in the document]
vinckr: Looked into it, not sure how to solve this in the wording. 
(You can use a username/pw model, but it needs an email as username?)

vinckr:
I caught some more errors and implemented your suggestions!

docs/docs/concepts/identity.mdx
```diff
@@ -12,19 +12,22 @@ check out the open source documentation:
 - [Identity Data Model](https://www.ory.sh/kratos/docs/concepts/identity-data-model)
 - [Identity Credentials](https://www.ory.sh/kratos/docs/concepts/credentials)
 
+The Identity Model can either be one of the available presets or a
+[customized model](#custom-identity-model).
+
 ## Available Presets
 
-At the moment we support three basic Identity Model Presets.
+The Ory Cloud provides three basic Identity Model Presets.
```
aeneasr, line 20:
```suggestion
Ory Cloud provides three basic Identity Model Presets.
```

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,293 @@
...
+    }
+```
+
+for example the GitHub Handle as `string` :
+
+```json
+"traits": {
+"type": "object",
+"properties": {
+        "handle": {
```
aeneasr, line 150: Incorrect indentation

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,293 @@
...
+      "required": [
+        "username"
+      ],
+      "additionalProperties": false
+    }
+```
+
+To add traits to the Identity Model, add them inside the `traits "properties"`
+
+```json
```
aeneasr, line 135: Use json5 if you include comments

aeneasr (changes requested):

docs/docs/guides/customize-identity-model.mdx
vinckr:
```suggestion
```json5
```
Do you mean this? 
It loses the color highlights if I use `json5` here. I think using `json` here, already renders it correctly.

with json5


with json

aeneasr: I think we just need to add json5 as an alias in the docs config. It would fix the markdown preview in github and potentially any CI we introduce in the future to validate markdown

vinckr: I changed it now in https://github.com/ory/docs/pull/453/commits/1ecf575728dede34d18a9ff64db96c5ee6194aaf to not require comments anymore. 
I think it is better this way. But we should still add json5 as alias. 
If you can give me some pointers, I could not figure out how to add it.

aeneasr (approved):

docs/docs/guides/customize-identity-model.mdx
```diff
@@ -0,0 +1,301 @@
+---
+id: customize-identity-model
+title: Customize Identity Model
```
aeneasr, line 3: Sorry, I just saw this - while it is named "Identity Model" we really should change it to "Custom Identity Schema". We will change this in all the UIs etc. 

So please replace `Identity Model` with `Identity Schema`!

aeneasr (changes requested):

docs/docs/guides/customize-identity-model.mdx
vinckr: I was going off this sentence in the identity concept docs:
`The Identity Model is a JSON Schema which describes your identity's (e.g. your customers / users / ...) fields.`

Do you think it would make sense to add the glossary from my other PR soon, for the different terms like Identity Model/Schema, Project Slug, etc: ?
https://github.com/ory/docs/blob/d0465fff31431b288ba9c1087a3757b7f160a6fc/docs/docs/concepts/self-service.mdx#glossary

I changed it, I hope it is clear in this document:
https://github.com/ory/docs/blob/a7438c42b2f0b837a9485d2a6d71c37ecc272953/docs/docs/concepts/identity.mdx

docs/docs/concepts/identity.mdx
```diff
@@ -6,25 +6,28 @@ title: Identity Data Model, Custom Fields
 The Identity Model is a JSON Schema which describes your identity's (e.g. your
 customers / users / ...) fields. In Ory, every identity can have their own
 model, allowing you to separate between different customer types, user types
-(e.g. employees and customers). For technical information on identity models,
+(e.g. employees and customers). For technical information on Identity Models,
```
aeneasr, line 9: Identity Schema

docs/docs/concepts/identity.mdx
```diff
@@ -6,25 +6,28 @@ title: Identity Data Model, Custom Fields
...
 check out the open source documentation:
 
 - [Identity Data Model](https://www.ory.sh/kratos/docs/concepts/identity-data-model)
```
aeneasr, line 12: We will need to rename this in the kratos docs. Please rename it here already

docs/docs/concepts/identity.mdx
```diff
@@ -82,3 +85,14 @@ field and additionally a first name, last name, and a checkbox.
   }
 }
 ```
+
+## Custom Identity Model
```
aeneasr, line 89: Schema

docs/docs/guides/customize-identity-schema.mdx
```diff
@@ -0,0 +1,302 @@
+---
+id: customize-identity-schema
+title: Customize Identity Schema
+---
+
+Ory supports custom Identity Models. To learn more about the concept of Identity
```
aeneasr, line 6: SCHEMA!

docs/docs/guides/customize-identity-schema.mdx
```diff
@@ -0,0 +1,302 @@
...
+Models at Ory, please head over to
```
aeneasr, line 7: Schema

aeneasr (changes requested):

aeneasr:
@vinckr please review PRs yourself by going through the diff ( https://github.com/ory/docs/pull/453/files ) before requesting reviews. It saves you time, the reviewer time, and gets things merged faster. Plus, reviewers will be much happier to review your things because they know that it can probably just pass.

vinckr:
Yes, I was aware that it is still named "identity model" in the concept docs. 
As I pointed out in an earlier comment:

```
I was going off this sentence in the identity concept docs:
The Identity Model is a JSON Schema which describes your identity's (e.g. your customers / users / ...) fields.
```

So the schema is just the technical device that describes the Identity Model, correct? 

Otherwise, as you commented, we need to change this in the OS documentation as well, as it is called "Identity Model" there. Imo it makes sense that the schema is describing the Identity Model.
So the schema is the practical thing you can change and the identity model is the abstract representation of identities in your system - the model.

> A schema is 
>  a diagrammatic presentation 
> a structured framework or plan : outline 

> A model is
> a pattern of something to be made
> an example for imitation or emulation
> structural design

vinckr:
Just to be clear: 

We are renaming "Identity Schema" into "Identity Model" in general, so the term "Identity Model" = "Identity Schema". Correct @aeneasr ?

aeneasr:
I think it would be much clearer if we use "Identity Schema" everywhere - including the OSS docs. Generally yes, the schema describes the model - but I think that we should use one terminology to not confuse people

vinckr:
Have changed all occurrences of Identity Model to Identity Schema. This PR should be ready now 🙂 .

PR to make the same changes for ory/kratos is almost ready. ory/meta and the API docs are following then.
I haven't changed the links here now, but I will link a PR here making the required changes to URLs then.
Sound good @aeneasr ?

To make the Ory specific terms crystal clear (also for other people writing docs) it would be good IMO to add a glossary:

> Do you think it would make sense to add the glossary from my other PR soon, for the different terms like Identity Model/Schema, Project Slug, etc: ?
> https://github.com/ory/docs/blob/d0465fff31431b288ba9c1087a3757b7f160a6fc/docs/docs/concepts/self-service.mdx#glossary

aeneasr:
Makes sense, thank you! :)

aeneasr (approved):

PR #457, opened 2021-08-11
fix: typo

marneysofia:
"gaining valuable insights in form of dashboards, and customization." > "and gaining
valuable insights in the form of dashboards and customization."
added "and" and  "the" and deleted an unnecessary comma.


aeneasr:
🎉

PR #458, opened 2021-08-11
fix: typo

marneysofia:
Added "and" at end of list:  "account recovery,
account setting, and account verification."
Verb-noun agreement: "Ory Cloud and our open source software is interlinked" > "Ory Cloud and our open source software are interlinked"
Re-write for flow: "The only impact Ory Cloud will have, is that development is accelerated as we are able to spend more resources on the open source!" > "The only impact Ory Cloud will have is that it will accelerate development and enable us to spend more resources
on the open source!"


aeneasr (approved):
Thank you again, @marneysofia :)

PR #459, opened 2021-08-11
fix: proxy command

vinckr:
The command doesn't work with capitalized P and I also get `Error: accepts 1 arg(s), received 3`
for `ory proxy local \  --port 4000 \  http://localhost:8000/`

It works with `ory proxy local --port 4000 http://localhost:8000/`

### DIFF
--- a/docs/docs/start-building/server-side-web-app.mdx
+++ b/docs/docs/start-building/server-side-web-app.mdx
@@ -80,9 +80,7 @@ Next, run the Ory Proxy with
   protecting.
 
 ```shell
-Ory Proxy local \
-  --port 4000 \
-  http://localhost:8000/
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you for the fix - however, the `\` are regular linux notation. It should work in any shell! But it probably does not work on Windows. So we can remove the `\` to make it compatible with windows too!

I overlooked this PR so please merge it with master (I fixed the casing issue in another PR). Thanks :)

vinckr:
Oh ok, the casing issue was the main thing for this PR.

It actually works in my shell now with the `\`, no idea why it didn't before. 

We can still merge this if you think its an improvement, otherwise feel free to close it.

PR #462, opened 2021-08-13
fix: adding notice to update CLI

vinckr:
As per https://github.com/ory-corp/cloud/issues/1352#issuecomment-897548305. 

How would the update work on Linux? Do you just run the install script again?


aeneasr:
> How would the update work on Linux? Do you just run the install script again?

Sorry missed this, yeah, absolutely!

PR #469, opened 2021-08-18
feat: services and apis

Benehiko:
fixes https://github.com/ory-corp/cloud/issues/1362


docs/docs/concepts/services-api.mdx
```diff
@@ -0,0 +1,13 @@
+---
+id: services-api
+title: Services & APIs
+---
+
+The Ory Cloud exposes APIs to its services to easily integrate your applications using our SDKs or
+HTTP Endpoints directly. The Ory Cloud APIs make it possible for you to integrate your own user interface (UI)
+in the form of a Web or Native Application.
```
vinckr, line 8: Is this possible now?
If not I would either omit it or refer to the GitHub issue/roadmap/relevant doc.

vinckr (changes requested):
I think you included some additional files by accident: 
` .idea/codeStyles/Project.xml ` etc.

You probably only want to add `services-api.mdx` and the .png, correct?

Benehiko:
Will this be adequate for the SDK urls page?

I wanted to add a link to an example application using the sdk's here, but if it's not necessary to do right now we can merge this as is.

aeneasr:
I think links to here would be good:

- https://www.ory.sh/docs/start-building
- https://www.ory.sh/docs/start-building/server-side-web-app
- https://www.ory.sh/docs/start-building/single-page-app-spa
- https://www.ory.sh/docs/start-building/native-mobile-app
- https://www.ory.sh/docs/start-building/other-languages

aeneasr:
Use the links you think make sense. We need to add the SDK initalization to those pages then!

aeneasr (approved):

PR #470, opened 2021-08-22
fix: typo

marneysofia:
"you want to distinct" -- "you want to distinguish" … distinct is adjective not verb.


aeneasr:
Ah yes, thank you!

PR #488, opened 2021-10-07
docs: add svelte node kratos

vinckr:
Adds `Svelte Node self service` to community projects

fyi @emrahcom


aeneasr:
Sorry I missed this - could you fix the conflict? TY

PR #492, opened 2021-10-10
updating the link for kratos-next due to username change

spa5k:
I changed my username, so thought it will be better if i change the link in here too


aeneasr:
thanks :)

PR #507, opened 2021-11-09
docs: early access

vinckr:
fyi @aeneasr @blueacidfrog
ready to merge 😬

### DIFF
--- a/docs/docs/get-started.mdx
+++ b/docs/docs/get-started.mdx
@@ -64,12 +64,11 @@ against it.
 _The Early Access period ends November 1st, 2022._
 
 After the early access period, you will be automatically notified about a
-suitable paying plan, determined by the current resource usage for each of your
-projects individually.
-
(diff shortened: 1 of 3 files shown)


docs/docs/get-started.mdx
```diff
@@ -1,44 +1,106 @@
...
+
+- Dedicated Developer Relations
+- Chat support via Slack channel
+- Ticket via GitHub Issues
+
+A ton of other features, including Ory Hydra and Ory Keto functionality, are
+[on the roadmap](./early-access.mdx) and will be worked on in the course of the
+Early Access.
+
+### Where is the catch
```
aeneasr, line 46: Maybe "What to expect" instead? Where's the catch is kinda negative

vinckr:
```suggestion
### What to expect
```

PR #508, opened 2021-11-09
docs: update to new ory cloud experience

aeneasr:


docs/docs/index.mdx
```diff
@@ -1,25 +1,68 @@
...
+  [Google's Zanzibar](https://research.google/pubs/pub48190/) research paper, an
+  infinitely scalable and blazing fast authorization and permission service.
+  Think RBAC on globally distributed steroids.
+
+The best place to run Ory software is in [Ory Cloud](https://www.ory.sh/). Ory
+Cloud takes care of scalability, reliability, latency, availability, security,
+and risk management for your authentication and authorization needs. Running Ory
+software on Ory Cloud is as easy as:
+
+```
```
Benehiko, line 36: Just some thoughts that I think might help with improving this section.

Could we maybe have a simple diagram here as well showing how each product fits in to the whole AuthN and AuthZ stack + the potential users' stack? This might also clarify immediately what people need.

We can also answer questions immediately through diagrams and simple language, such as:
- Is the Cloud offering just Kratos? 
- Is the cloud all of these listed? 
- Or is the cloud a separate thing?
- Can I implement my own branding?
- Is it "compatible" with JavaScript? (dumb question, but people might not fully get "api only" or "api first". We might need to make that more clear.
- Do I need to pay anything to start using these?

vinckr:
I think 
https://github.com/ory/docs/blob/dc078262b2cb86dad3ef1a4a4fefb9f899e2186d/docs/docs/index.mdx
is a pretty good overall introduction. 
The current "welcome to Ory" is a bit drawn out IMO and this would be more compressed.

From the top of my head, this is how I would improve the "welcome" page:

1.
Add https://github.com/ory/docs/blob/dc078262b2cb86dad3ef1a4a4fefb9f899e2186d/docs/docs/index.mdx
as intro text instead of the current.

2.
remove https://www.ory.sh/docs/welcome#ory-cloud 
- its frankly a bit confusing and party redundant information ("Ory Cloud uses a variety of components")
- it places too much importance on the managed UI
- it makes it seems Ory Cloud is just Kratos / the other text is much more intuitive and sums it up (aka "These are Ory services -> most convenient way to run them is in Ory Cloud")

3.
https://www.ory.sh/docs/welcome#ory-open-source
this section could be cut down a bit, basically 2 paragraphs instead of 5

4.
https://www.ory.sh/docs/welcome#new-to-ory-
This can be moved to the website IMO, as a general "new to ory" intro page for developers or as part of the community / learn pages.

---

Generally we could move most of this "welcome to Ory page" to the website.

PR #515, opened 2021-11-24
fix: tac PR projects doc

vinckr:
@tacurran 
I renamed the file again, then we don't have to change the sidebar, redirects and so on.

Title of the file is still "Products".


docs/docs/ecosystem/projects.md
```diff
@@ -0,0 +1,86 @@
+---
+id: projects
+title: Products
```
aeneasr, line 3:
```suggestion
id: products
title: Projects
```

PR #522, opened 2021-12-07
fix: cloud proxy docs

vinckr:
closes https://github.com/ory-corp/cloud/issues/1684

- adds ORY_SDK_URL to terminology doc
- puts Ory specific terms at the top of said doc
- removes ory proxy local as the command is no longer working
- shortens headings from "Ory Proxy for Local Development" to "Local Development", looks cleaner, and should be clear from the heading that it concerns ory proxy. 

optional can be reverted:
- removes the articles, "The Ory Proxy" -> "Ory Proxy", which makes it sound like a proper product. if merged, will remove other articles as well.

### DIFF
--- a/docs/docs/guides/proxy.mdx
+++ b/docs/docs/guides/proxy.mdx
@@ -1,13 +1,13 @@
 ---
 id: proxy
-title: Using the Ory Proxy
+title: Using Ory Proxy
 ---
 
-The Ory Proxy is a powerful tool which authenticates users, converts session
+Ory Proxy is a powerful tool which authenticates users, converts session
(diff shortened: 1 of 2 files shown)


docs/docs/guides/proxy.mdx
```diff
@@ -171,28 +170,30 @@ Cookie:
 }
 ```
 
-## Ory Proxy for Local Development
+## Local Development
```
aeneasr, line 173: Keeping names explicit in the headlines is good for SEO. If someone searches "ory proxy" on google, it's more likely that they'll end up on this page!

vinckr: Makes sense, reverted it!

PR #528, opened 2022-01-03
fix: clarify managed UI location

vinckr:
I could not really find the managed UI after setting up my CNAME, this will hopefully prevent others from searching :-).

### DIFF
--- a/docs/docs/guides/custom-domains.mdx
+++ b/docs/docs/guides/custom-domains.mdx
@@ -55,6 +55,9 @@ section:
 
 When the domain is reachable, it will have the "Active status".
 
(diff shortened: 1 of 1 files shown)


aeneasr:
It’s on the connect page :)

vinckr:
True, but on the connect page it does not use the CNAME. 
It is currently not evident in the UI where i can use the managed UI with CNAME, as in 
https://auth.vinckr.com/ui/welcome
opposed to 
https://distracted-elion-yl0zho4q9i.projects.oryapis.com/ui/welcome

It is pretty self-evident, when you see it, but I was not sure if the managed UI would even work with CNAME.

aeneasr:
Got it, this is an issue for ory/cloud & michael. Can you please create an issue in ory-corp/cloud and tag him?

PR #540, opened 2022-01-10
feat: bug bounty doc

vinckr:
Adds the bug bounty document to the OS ecosystem documentation. 

We can point to it at an appropriate place in cloud docs. 

When we have the platform for payouts ready we can point to it in this document, but for now it does a good job of fulfilling its purpose:
>Define the submission and follow-up process of reporting security issues to Ory.

tagging you for review, since you wrote it zepatrik

### DIFF
--- a/docs/docs/ecosystem/bugbounty.md
+++ b/docs/docs/ecosystem/bugbounty.md
@@ -1,0 +1,55 @@
+---
+id: bugbounty
+title: Bug Bounty Program
+---
+
+As a security focused company, Ory appreciates and encourages any interest by
+security researches and white-hat hackers in the companies products. To support
+the global infosec community, we award the responsible disclosure of
+vulnerabilities and data leaks according to the following policies.
+
+## Responsible Disclosure
+
+A responsible disclosure always starts with contacting the Ory security team,
+using one of the channels specified at
+[https://www.ory.sh/.well-known/security.txt](https://www.ory.sh/.well-known/security.txt).
+Provide all details to reproduce the issue in form of a non-public git
+repository including all setup and execution steps, the binary and source of
+the exploit, as well as an example exploit payload (if applicable). Low-quality
+reports may be disqualified. The Ory security team will contact you within 96
+hours to confirm the issue or ask for further details. Do not publish or share
+any details regarding the issue until the Ory security team explicitly permits
+the publication in written form. The Ory security team will permit publication
+once incidence response and patches are fully enrolled. Any exploits shall not
+be tested against environments hosted by Ory or any third party without the
+explicit consent to do so. In case this happens, the submission is disqualified.
+
+## Awards
+
+The Ory maintainers have final decision on which issues constitute security
+vulnerabilities. Reports for already known issues will only be awarded with
+regards to the new information they add to the issue. Ory will score the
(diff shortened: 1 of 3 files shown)


zepatrik (changes requested):
Generally this looks good to me :+1:

docs/docs/ecosystem/bugbounty.md
```diff
@@ -0,0 +1,55 @@
+---
+id: bugbounty
+title: Bug Bounty Program
+---
+
```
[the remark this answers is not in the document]
zepatrik: Ah wait, this is docusaurus. Use admonitions then:

```suggestion

:::caution Disclaimer

We are currently in the process of setting up a disclosure platform. Until that process is done, we cannot payout bug bounties due to the administrative overhead. All disclosures that are made until the launch, will later be processed under the same conditions, but we cannot guarantee when that happens.

:::

```

docs/docs/ecosystem/bugbounty.md
```diff
@@ -0,0 +1,61 @@
+---
+id: bugbounty
+title: Bug Bounty Program
```
aeneasr, line 3: Maybe "Responsible Disclosure" or "Security Policy"?

docs/docs/ecosystem/bugbounty.md
```diff
@@ -0,0 +1,61 @@
...
+---
+
+:::caution Disclaimer
+
+We are currently in the process of setting up a disclosure platform. Until that process is done, we cannot payout bug bounties due to the administrative overhead. All disclosures that are made until the launch, will later be processed under the same conditions, but we cannot guarantee when that happens.
+
+:::
+
```
aeneasr, line 11: How does 

https://github.com/ory/kratos/security/policy

in particular "Supported Versions" fit into this here? I think we could use this document here also for the security policy in the open source?

aeneasr (commented):
Looking great!

docs/docs/ecosystem/bugbounty.md
vinckr: Hm this was mostly for Ory Cloud, but the same policy should apply to OS.
In any case, I think we need the SECURITY.md or GitHub complains, but we could copy its contents to this document and then just link the /docs in SECURITY.md

Supported Versions does not make a lot of sense in the Ory Cloud context, but if we explain it briefly (underlying core services etc.) it should work.

wdyt?

docs/sidebar.json
```diff
@@ -66,6 +66,7 @@
       "ecosystem/logging",
       "ecosystem/contributing",
       "ecosystem/sqa",
+      "ecosystem/bugbounty",
```
harnash, line 69: @vinckr I think this breaks Kratos docs bulds:
```json
Unable to build website for locale "en".
Error: Server responded to https://raw.githubusercontent.com/ory/docs/master/docs/docs/ecosystem/bugbounty.md with status code 404:
404: Not Found
```

zepatrik: True, the sidebar has the wrong ID. It should be
```suggestion
      "ecosystem/security",
```

aeneasr:
Thanks, resolved

PR #548, opened 2022-01-21
fix: correct npm package in typescript example

vinckr:
changes the typescript expressjs example to use @ory/client instead of @ory/kratos-client. 
tested the example and it works just as before.


aeneasr:
Did you test if this works? The two SDKs are not full ycompatible

vinckr:
the guide at https://www.ory.sh/docs/start-building/server-side-web-app#expressjs--nodejs
works fine
Screenshot 2022-01-21 at 19 25 22

PR #552, opened 2022-02-02
docs: ory cloud password policy

vinckr:
It is pretty similar to https://www.ory.sh/kratos/docs/next/concepts/security/#password-policy

But it describes the currently implemented password policy in Ory Cloud.

### DIFF
--- a/docs/docs/concepts/password-policy.mdx
+++ b/docs/docs/concepts/password-policy.mdx
@@ -5,12 +5,12 @@ title: Password Policy
 
 Password-based authentication flows are subject to frequent abuse through social
 engineering, password guessing and phishing attacks.  
-Ory Cloud implements measures to provide high security for password-based
-flows. The Ory Cloud password policy follows standards by the _National Cyber Security
+Ory Cloud implements measures to provide high security for password-based flows.
+The Ory Cloud password policy follows standards by the _National Cyber Security
 Centre_
 ([`NCSC`](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach))
-and _National Institute of Standards and
-Technology_ ([`NIST`](https://pages.nist.gov/800-63-3/sp800-63b.html)) as well as
+and _National Institute of Standards and Technology_
+([`NIST`](https://pages.nist.gov/800-63-3/sp800-63b.html)) as well as
 [leading](https://www.troyhunt.com/passwords-evolved-authentication-guidance-for-the-modern-era/)
 security researchers.
 
--- a/docs/docs/concepts/password-policy.mdx
+++ b/docs/docs/concepts/password-policy.mdx
@@ -6,30 +6,35 @@ title: Password Policy
 Password-based authentication flows are subject to frequent abuse through social
 engineering, password guessing and phishing attacks.  
 Ory Cloud implements measures to provide high security for password-based
-flows.  
-The Ory Cloud password policy is based on recommendations by the
-[`NIST`](https://pages.nist.gov/800-63-3/sp800-63b.html) and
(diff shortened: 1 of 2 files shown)


docs/docs/concepts/password-policy.mdx
```diff
@@ -0,0 +1,49 @@
...
+engineering, password guessing and phishing attacks.  
+Ory Cloud implements measures to provide high security for password-based
+flows.  
+The Ory Cloud password policy is based on recommendations by the
+[`NIST`](https://pages.nist.gov/800-63-3/sp800-63b.html) and
+[`NCSC`](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach)
+institutes among
+[others](https://www.troyhunt.com/passwords-evolved-authentication-guidance-for-the-modern-era/).
+
+- The password must at least be 8 characters long and all characters (unicode,
```
aeneasr, line 16: This is actually configurable now :) But it is 8 per default

docs/docs/concepts/password-policy.mdx
```diff
@@ -0,0 +1,49 @@
...
+  ASCII) must be allowed.
+- Ory Cloud makes sure the password is not similar to the username/email or
+  other credentials.  
+  To ensure the password is different, Ory Cloud enforces a minimum
+  [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance)
+  (measuring how much two strings resemble each other). It also makes sure no
+  significant strings of the credentials are part of the password.
```
aeneasr, line 23: Mabye give an example:

> Password `angel12` would not be valid for someone with a username of `angel`

docs/docs/concepts/password-policy.mdx
```diff
@@ -0,0 +1,49 @@
...
+- Ory Cloud checks all passwords against the
+  [`HIBP`](https://haveibeenpwned.com/) API.  
+  `HIBP` provides a open database of leaked passwords. Every password sent to
+  `HIBP` is SHA-1 hashed and anonymized, so there is no compromise of user
```
aeneasr, line 27: I suggest to focus on the feature rather than how we achieve it. So something like: "Ory checks all passwords in an anonymized way against a database of known leaked passwords using the Have I Been Pwned API.

docs/docs/concepts/password-policy.mdx
```diff
@@ -0,0 +1,49 @@
...
+  credentials.
+- Ory Cloud does not require or prohibit a mixture or repeated characters.
```
aeneasr, line 29: Explain why or have a reference to somewhere

aeneasr (commented):
good job!

docs/docs/concepts/password-policy.mdx
vinckr: I include something about the default and we mention that you can change it in this doc when it is documented how to configure it.

docs/docs/concepts/password-policy.mdx
```diff
@@ -0,0 +1,54 @@
...
+  other credentials.  
+  To ensure the password is different, Ory Cloud enforces a minimum
+  [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance). It
+  also makes sure no significant strings of the credentials are part of the
+  password. For example if an users email is `bob@example.com`, `bob24` would
+  not be a valid password.
+- Ory Cloud checks all passwords against a database of known leaked passwords
+  through the [`HIBP`](https://haveibeenpwned.com/) API.  
+  Breached or leaked password detection uses anonymized data.
+
```
aeneasr, line 31:
```suggestion
```

PR #571, opened 2022-02-10
feat: configuration json schema live editor

aeneasr:
Thank you for the idea @vinckr @Benehiko 

I think this is not the best solution ever, but it's a bit better than just the text file. @vinckr looks like we need to add a lot of "title" and "description" values to the config items for the structure to look nicer :)

Maybe we can also add some type of YAML output of the current config somewhere...

Screenshot 2022-02-10 at 23 42 10

### DIFF
--- a/docs/hydra/reference/config.schema.json
+++ b/docs/hydra/reference/config.schema.json
@@ -1,0 +1,1057 @@
+{
+  "$id": "https://github.com/ory/hydra/docs/config.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Ory Hydra Configuration",
+  "type": "object",
+  "definitions": {
+    "http_method": {
+      "type": "string",
+      "enum": [
+        "POST",
+        "GET",
+        "PUT",
+        "PATCH",
+        "DELETE",
+        "CONNECT",
+        "HEAD",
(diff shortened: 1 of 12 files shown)


Benehiko:
> Thank you for the idea @vinckr @Benehiko
> 
> I think this is not the best solution ever, but it's a bit better than just the text file. @vinckr looks like we need to add a lot of "title" and "description" values to the config items for the structure to look nicer :)
> 
> Maybe we can also add some type of YAML output of the current config somewhere...
> 
> Screenshot 2022-02-10 at 23 42 10

Yeah, there are lots of missing descriptions which creates problems in understanding the config by just seeing it here, we probably would need to have more references here to the documentation or something to make it more legible. 

Having a live editor that can switch between YAML and JSON would be perfect, but of course, we can start off with just a simple "export into filename.yaml/json"

vinckr:
I will personally use this, love it 😍 
I can take this on and see that we have proper descriptions etc - unless there is something major with the website coming up for me, still unclear.

aeneasr:
Great! I think we can easily add the descriptions. Please add the descriptions in the repositories themselves, so for example: https://github.com/ory/kratos/blob/master/embedx/config.schema.json

If you don't find the file LMK

PR #588, opened 2022-02-14
feat: vale linter action

vinckr:
Adds the Vale Linter Action to improve consistency of style across the documentation.

It currently takes the write-good and Microsoft styles, both have reasonable suggestions. We could also add our own styles, or modify the existing.

https://github.com/vinckr/vale/
https://github.com/marketplace/actions/vale-linter

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -18,7 +18,7 @@ jobs:
           styles: |
             https://github.com/errata-ai/Microsoft/releases/latest/download/Microsoft.zip
             https://github.com/errata-ai/write-good/releases/latest/download/write-good.zip
-          config: https://raw.githubusercontent.com/ory/docs/master/.vale.ini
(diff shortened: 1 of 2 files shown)


.github/workflows/vale.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Lint Markdown with Vale
+        uses: errata-ai/vale-action@master
+        with:
+          styles: |
+            https://github.com/errata-ai/Microsoft/releases/latest/download/Microsoft.zip
+            https://github.com/errata-ai/write-good/releases/latest/download/write-good.zip
+          config: https://raw.githubusercontent.com/ory/docs/e66b0701ef8c4d18c32d2488ac905cb54e23f581/docs/.vale.ini
```
aeneasr, line 21: You can probably `${{ github.sha }}` here

vinckr: works great thanks.

PR #601, opened 2022-02-16
feat: added documentation about sending sms with courier

alexey-reshetnik:


docs/kratos/concepts/email-sms.md
```diff
@@ -133,6 +134,112 @@ courier:
...
+  sms:
+    from: '+12065550110'
+```
+
+#### Request configuration
+
+```yaml title="path/to/my/kratos/config.yml"
+# $ kratos -c path/to/my/kratos/config.yml serve
+courier:
+  sms:
```
aeneasr, line 183: Since this is so generic now, would it make sense to just call it the HTTP-RPC courier? And that can be used for sending SMS, but could also be wired to e.g. some CRM?

Or is there still some magic that happens that is required for SMS?

alexey-reshetnik: Yes, there are templates defined for SMS, so this integration can't be reused for more generic request sending. I think that it would be better to create a dedicated method in courier that will allow sending just plain HTTP requests

Issue #532, opened 2022-01-06
Document use of Go SDK


aeneasr:
https://ory-community.slack.com/archives/C02MR4DEEGH/p1641468182003000

vinckr:
What is the objective of this issue? 

I saved this code example but it is using kratos-client; should it be the "new" client now? 
```
package ory

import (
	"context"
	"errors"

	ory "github.com/ory/kratos-client-go"
)

type CreateIdentityParams struct {
	Email    string
	Password string
}

// CreateIdentityFlow create an identity with the provided password
func (c *Client) CreateIdentityFlow(ctx context.Context, params CreateIdentityParams) (string, error) {
	flow, _, err := c.oryhttpc.V0alpha2Api.InitializeSelfServiceRegistrationFlowWithoutBrowser(ctx).Execute()
	if err != nil {
		return "", err
	}

	result, res, err := c.oryhttpc.V0alpha2Api.SubmitSelfServiceRegistrationFlow(ctx).Flow(flow.Id).SubmitSelfServiceRegistrationFlowBody(
		ory.SubmitSelfServiceRegistrationFlowWithPasswordMethodBodyAsSubmitSelfServiceRegistrationFlowBody(&ory.SubmitSelfServiceRegistrationFlowWithPasswordMethodBody{
			Method:   "password",
			Password: params.Password,
			Traits: map[string]interface{}{
				"email": params.Email,
			},
		}),
	).Execute()

	defer res.Body.Close()

	if err != nil {
		c.logger.Error(err, "")
		return "", err
	}

	if result.Session == nil {
		return "", errors.New("server did not create sessions for new registration")
	}

	return result.Identity.Id, nil
}
```

Also created a draft PR, but it is not correct yet. 


I could modify this example from Alano for locally run Kratos wdyt? 
https://www.ory.sh/docs/guides/protect-page-login/go

aeneasr:
yes, it should be the new client :)

PR #608, opened 2022-02-21
docs: draft go sdk doc

vinckr:
draft for https://github.com/ory/docs/issues/532

### DIFF
--- a/docs/kratos/sdk/go.md
+++ b/docs/kratos/sdk/go.md
@@ -18,6 +18,12 @@ We use code generation to generate our SDKs. The Go SDK is generated using
 import "github.com/ory/client-go/client"
 ```
 
+To use a proxy, set the environment variable HTTP_PROXY:
+
+```go
+os.Setenv("HTTP_PROXY", "http://proxy_name:proxy_port")
+```
+
 ## Make requests
 
 Making requests is straight forward:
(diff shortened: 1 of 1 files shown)


aeneasr:
Great! This is going in a good direction!

vinckr:
I "fixed" the examples, but I am not sure if they are 100% correct, due to lack of experience with Go.

docs/kratos/sdk/go.md
```diff
@@ -0,0 +1,126 @@
...
+
+```go
+import "github.com/ory/client-go/client"
+```
+
+To use a proxy, set the environment variable HTTP_PROXY:
+
+```go
+os.Setenv("HTTP_PROXY", "http://proxy_name:proxy_port")
+```
```
Benehiko, line 25: So the client will read the `HTTP_Proxy` env variable by itself?

docs/kratos/sdk/go.md
```diff
@@ -0,0 +1,126 @@
...
+func main() {
+    KratosPublicEndpoint := url.Parse("https://kratos.localhost:4433")
+
+    flow, _, err := s.KratosPublicEndpoint.V0alpha2Api.GetSelfServiceLoginFlow(r.Context()).Id(flowID).Cookie(cookie).Execute()
+    if err != nil {
+        writeError(w, http.StatusUnauthorized, err)
+        return
+    }
+}
+```
```
Benehiko, line 47: where does `s` come from? also why declare `KratosPublicEndpoint` when it isn't used?

docs/kratos/sdk/go.md
```diff
@@ -0,0 +1,126 @@
...
+func (c *Client) DeleteIdentity(ctx context.Context, identityID string) error {
+	res, err := c.oryhttpc.V0alpha2Api.AdminDeleteIdentity(c.contextWithSecret(ctx), identityID).Execute()
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	return nil
+}
+```
```
Benehiko, line 72: `ctx` isn't used anywhere in its scope. then you have a function `DeleteIdentity` which is doing the same thing more or less to the code above? I think this needs a bit of cleanup. Also where does `contextWithSecret` come from?

Reading this makes me confused on how to use Authorization with the ory client.

docs/kratos/sdk/go.md
```diff
@@ -0,0 +1,126 @@
...
+For more information on Authorization, check the
+[go-swagger documentation](https://goswagger.io/generate/client.html#authentication).
+
+## Create Identity
+
+Code example for deleting an identity using the PAT to interact with Kratos via the admin URL:
```
Benehiko, line 78: I thought we are creating an identity not deleting :upside_down_face:

docs/kratos/sdk/go.md
```diff
@@ -0,0 +1,126 @@
...
+		c.logger.Error(err, "")
+		return "", err
+	}
+
+	if result.Session == nil {
+		return "", errors.New("Server did not create a session for the new registration.")
+	}
+
+	return result.Identity.Id, nil
+}
```
Benehiko, line 124: I would stay away from functions and structs bound to functions since this adds specific logic which has nothing to do with the SDK. Also these steps do not make sense if you are creating an identity through the admin api? (i thought that was what it's supposed to do because of the delete through the admin api). 

In general you would most likely use redirects in Go but can also (in cases of say a CLI) do a flow initialisation without a browser like you are doing here. But then make sure that the code follows logically.

This code doesn't even close its brackets.

`result, res, err := c.oryhttpc.V0alpha2Api.SubmitSelfServiceRegistrationFlow(ctx).Flow(flow.Id).SubmitSelfServiceRegistrationFlowBody(`

Benehiko (changes requested):
I would suggest building a Go example and then copy paste or reference the code in the markdown with the proper formatting. Also no need to go complicated with structs and functions, keep it simple as possible with no extra logic required to understand the code.

Also always show where something is being declared when used later, such as the client is used but where are you getting it from? Comment in the code what it's doing to ensure readability. 

Ensure to start off with creating an account before deleting an account, so make sure the order of things are correct. Distinguish between admin and public flows in your examples as well and make it clear for what purpose in which use case you would use such a function. 

E.g. when you have a browser you probably want to do redirects. When you have a CLI you don't have a browser. When you have a server you probably want to do Admin requests.

PR #616, opened 2022-02-24
feat: add php guide

adamwalach:
### Changes proposed in this pull request:
- added PHP guide for "Protect a Page or Screen with Login"
- updated tests to include PHP case
- parts of text common for multiple languages moved to templates


docs/guides/protect-page-login/php.mdx
```diff
@@ -0,0 +1,135 @@
...
+import mp4 from './go/screencast.mp4'
+import webm from './go/screencast.webm'
+
+<Teaser
+  framework={<a href="https://www.php.net/">PHP</a>}
+  mp4={mp4}
+  webm={webm}
+/>
+```
+
```
Demonsthere, line 18: Please add a small introduction paragraph like we do here https://www.ory.sh/docs/guides/protect-page-login/expressjs

Demonsthere (changes requested):

Demonsthere (approved):
From a documentation/language point of view: lgtm :)

docs/guides/protect-page-login/php.mdx
```diff
@@ -0,0 +1,135 @@
...
+```shell-session
+composer require ory/client:^v0.0.1-alpha
+```
+
+To simplify URLs handling we install the [bramus/router](https://packagist.org/packages/bramus/router)
+```shell-session
+composer require bramus/router
+```
+
+### Why do I Need the Ory CLI?
```
zepatrik, line 42: Is this shared between all guides? Can we make it an importable component then, so that we only have to fix it in one place later on?

docs/guides/protect-page-login/php.mdx
```diff
@@ -0,0 +1,135 @@
...
+<SdkEnvVar />
+```
+
+```shell-session
+ory proxy http://localhost:3000
+```
+
+Then open [http://localhost:4000](http://localhost:4000) in your browser. You
+are presented with Ory's Sign In page! Let's click on sign up and create your
+first user!
```
zepatrik, line 116: I find it a bit confusing to start the server on `:3000` previously, but then going to `localhost:4000`. Can you maybe add something the lines of

> ... access the php app through the proxy ...

zepatrik (commented):
Nice, just a bit of text improvements. Don't know php, but it looks fairly simple :+1:

docs/guides/protect-page-login/php.mdx
adamwalach: good idea, I'll move this section to the `_common` dir

docs/guides/protect-page-login/php.mdx
adamwalach: right, I changed it to:
```To access the PHP app through the ORY proxy open```

zepatrik (approved):
Very nice, just one last thing is to run `npm run format` :wink:

PR #617, opened 2022-02-24
docs: add identity import

aeneasr:


docs/guides/manage-identities/account-recovery.mdx
```diff
@@ -0,0 +1,187 @@
...
+  </TabItem>
+  <TabItem value="kratos">
+    <SelfHostedKratos />
+  </TabItem>
+</Tabs>
+````
+
+The response contains a `recovery_link` value which is the link the user should
+use to set up his or her credentials (e.g. connect to a Social Sign In Provider,
+set up a password, ...):
```
vinckr, line 85:
```suggestion
The response contains a `recovery_link` value. This is the link the user should
use to set up his or her credentials to connect to a Social Sign In Provider or
set up a password :
```

Again as above we have to decide how we spell "Social Sign In Provider" to make it consistent.

Also latin abbrevations and parantheses.

docs/guides/manage-identities/account-recovery.mdx
```diff
@@ -0,0 +1,187 @@
...
+needs to be issued and the user needs to re-do the flow.
+
+It is currently not possible to send the recovery link directly to a user's
+email, this feature is tracked as
+[#595](https://github.com/ory/kratos/issues/595).
+
+### Enable Account Recovery
+
+To enable recovery flows, make the following adjustments to your Ory Kratos
+configuration:
```
vinckr, line 132:
```suggestion
To enable recovery flows, make the following adjustments to your Ory Kratos
configuration:
```

This will probably be covered in different documentation, but right now it is very unclear (from this document alone) how this is to be actually done in Ory Cloud.

vinckr (changes requested):
I have a few comments, 
some can be safely ignored / are more a matter of taste.

docs/guides/manage-identities/account-recovery.mdx
aeneasr: You're right, could you please add some screen shots to explain this a bit in a follow up PR?

PR #624, opened 2022-02-27
docs(kratos): non-blocking web hooks

aeneasr:
See https://github.com/ory/kratos/pull/2266

### DIFF
--- a/docs/kratos/self-service/hooks.mdx
+++ b/docs/kratos/self-service/hooks.mdx
@@ -139,6 +139,22 @@ object into the jsonnet template as well. These objects are available through a
 send to the web hook end point, the jsonnet template would look like this:
 `function(ctx) { user_id: ctx.identity.id }`
 
+#### Non-Blocking Web Hooks
+
+Sometimes you need to notify a system without needing its response. For example,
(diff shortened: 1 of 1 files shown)


docs/kratos/self-service/hooks.mdx
```diff
@@ -139,6 +139,22 @@ object into the jsonnet template as well. These objects are available through a
...
 `function(ctx) { user_id: ctx.identity.id }`
 
+#### Non-Blocking Web Hooks
+
+Sometimes you need to notify a system without needing its response. For example,
+when a user signs up and you want their email address to be added to the
+newsletter you might not care about the response and favor the execution to be
+non-blocking.
+
+For those cases you can set `response.ignore` to `true` in your web-hook config:
```
tomekpapiernik, line 149:
```suggestion
Sometimes you need to notify a system without needing its response. For example,
when a user signs up, their email address is added to the newsletter. If the system's response
is not important, you can ignore the response and make the webhook execution non-blocking.

To do that, set `response.ignore` to `true` in the webhook config:
```

tomekpapiernik (changes requested):

PR #654, opened 2022-03-07
docs(kratos): passwordless webauthn

aeneasr:
## TO DO

* [ ] Write "try the flow" guide in webauthn document
* [ ] Write config docs for ory cloud


tomekpapiernik (changes requested):
IMO you're overusing imports. The things you're writing about are
- very specific
- described for the very first time
- they are not reused multiple times across different docs

For now, I think that these docs would be easier to maintain to shape without imports. If, in the future, it turns out that we need to reuse certain sections of these docs in multiple places in our docs, then we make these sections "importable".

aeneasr:
This PR adds three new pages:

- https://docs-oantiiapm-ory.vercel.app/docs/kratos/concepts/credentials/webauthn
-https://docs-oantiiapm-ory.vercel.app/docs/guides/passwordless/webauthn 
- https://docs-oantiiapm-ory.vercel.app/docs/kratos/guides/two-factor-authentication-2fa-mfa#fido2--u2f-webauthn

They reuse the same components, which might be a bit repetitive, but it's always useful information.

PR #668, opened 2022-03-14
docs: update redirects doc

vinckr:
Make the redirection document more clear as described here:
closes https://github.com/ory/cloud/issues/51
Also revised the entire document

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).


aeneasr (approved):

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
-page. To return the user back to the original url we can append
-`?return_to=https://myapp.com/posts` to the specified Self-Service initialise
-flow url.
-
-The `Allow list` is used in this scenario to prevent Open Redirect Attacks by
-only allowing certain domains, or paths on a domain. As shown in this example
-below the https://sub.domain.myapp.com/only/path needs to match perfectly for
-this sub-domain and path for it to be allowed. Other redirects using myapp.com
-will fail, however, for https://anotherapp.com will succeed on any path.
+parameter on [Self-Service Flows](./self-service.mdx). For example: A user opens
```
tomekpapiernik, line 18:
```suggestion
parameter on [Self-Service Flows](./self-service.mdx). For example: a user opens
```

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
+a sharable link to go to `https://myapp.com/posts`. This URL requires the user
+to be signed in and redirects the user back to the login page. To return the
+user back to the original URL append `?return_to=https://myapp.com/posts` when
```
tomekpapiernik, line 21:
```suggestion
user back to the original URL, append `?return_to=https://myapp.com/posts` when
```

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
 
-The Login and Registration flows have three fields (all optional):
-
-1. `Default Redirect URL` domain or path relative to your application URL
-2. An `Post-Password Redirect URL`
-3. An `Post-OIDC Redirect URL` (**not supported**)
+### Login, registration and settings
 
-And Settings has three fields (all optional):
+The `Login` and `Registration` flows have three fields (all optional):
```
tomekpapiernik, line 54: Please get rid of the code formatting for Login and Registration. Previously the flow names were in bold so let's be consistent and bold them here, too. (Not a fan of that too, but at least we'll be consistent ;) )

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
 
 1. `Default Redirect URL` domain or path relative to your application URL
-2. An `Post-Password Redirect URL`
-3. An `Post-Profile Redirect URL`
+2. `Post-Password Redirect URL`
+3. `Post-OIDC Redirect URL` (_coming soon_)
```
tomekpapiernik, line 58: No "coming soon" in the docs please. Docs should describe the current state of the software.

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
-https://myapp.com/after/login/password.
+In the example below , the `Post-Login Default Redirect URL` has been set to use
+a FQDN. This will set the Login `Default Redirect URL` to
+`https://myapp.com/after/login` and the `Post-Password` to
+`https://myapp.com/after/login/password`.
 
 ![Browser Redirects URL Management Transformed](/images/browser-redirects/browser-redirects-url-management-2.png)
 
-### Verification, Recovery and Logout
+### Verification, recovery and logout
```
tomekpapiernik, line 103:
```suggestion
### Verification, recovery, and logout
```

docs/concepts/redirects.mdx
```diff
@@ -12,133 +12,156 @@ to configure it on Ory Cloud console. If you are interested in utilizing the SDK
...
 
-![Browser Redirects URL Error Message](/images/browser-redirects/browser-redirects-error-messages-1.png)
+```ts
+          .then((res) => {
+            router.push('/<your-route>')
+          })
+```
 
-### Post Flow
+### `incorrect format`
```
tomekpapiernik, line 146: Let's avoid code in headings. Sometimes you just can't, but here your heading is JUST `code`. Try to think of a better one ;)

tomekpapiernik (dismissed):

vinckr:
Need to fix a few things, ETA 30 min.

docs/concepts/redirects.mdx
```diff
@@ -3,35 +3,33 @@ id: redirects
...
+
+Set dynamic redirects using the `?return_to=` query parameter on
+[Self-Service Flows](./self-service.mdx). For example: a user opens a sharable
+link to go to `https://myapp.com/posts`. This URL requires the user to have an
+active session and redirects the user back to the login page. To return the user
+back to the original URL, append `?return_to=https://myapp.com/posts` when
+starting the self-service login flow `initSelfServiceLogin`:
+
+```js
+    "request_url": "http://<your-project>.projects.oryapis.com/self-service/login/browser?return_to=...",
```
aeneasr, line 20: This is confusing, `initSelfServiceLogin`  does not accept the aprameter like this. It also depends on the SDK that you're using. Maybe it makes more sense to link to the API doc here:

https://www.ory.sh/docs/reference/api#operation/submitSelfServiceLoginFlow

and show an example with cURL:

```suggestion
curl -X GET 'http://<your-project>.projects.oryapis.com/self-service/login/browser?return_to=...'
```

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,125 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
+2. `Post-Password Redirect URL`
+3. `Post-Profile Redirect URL`
+
+To overwrite the `Global Redirect URL` set the `Default Redirect URL` to a fully
+qualified domain name
+([FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) - for example
+`https://domain.example/login`).  
+To replace the `Global Redirect URL` path set the `Default Redirect URL` to a
+path. For example when setting it to `/login`,
+`https://domain.example/global/path` becomes `https://domain.example/login`.
```
aeneasr, line 67: I don't think this works any more, can you confirm that this is true? I think you now just need to always specify the FQDN if it's not your CNAME or hosted URL or Ory Proxy base URL

aeneasr (changes requested):

aeneasr:
Also, I have to say that I find this quite difficult to read, because there are so many code boxes:

Bildschirmfoto 2022-03-18 um 14 26 57

Can there be maybe a bit better format? Code boxes should really only be used for code.

Another question is whether it would make sense to add a mermaid graph explaining the logic how we choose which method gets used?

docs/concepts/redirects.mdx
```diff
@@ -3,35 +3,33 @@ id: redirects
...
+
+Set dynamic redirects using the `?return_to=` query parameter on
+[Self-Service Flows](./self-service.mdx). For example: a user opens a sharable
+link to go to `https://myapp.com/posts`. This URL requires the user to have an
+active session and redirects the user back to the login page. To return the user
+back to the original URL, append `?return_to=https://myapp.com/posts` when
+starting the [self-service login flow](https://www.ory.sh/docs/reference/api#operation/initializeSelfServiceLoginFlowForBrowsers):
+
+```js
+    curl -X GET 'http://<your-project>.projects.oryapis.com/self-service/login/browser?return_to=...'
```
aeneasr, line 20:
```suggestion
curl -X GET 'http://<your-project>.projects.oryapis.com/self-service/login/browser?return_to=...'
```

aeneasr (changes requested):

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,125 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
+2. Post-Password Redirect URL
+3. Post-OIDC Redirect URL
+
+The Settings flow has three fields (all optional):
+
+1. Default Redirect URL domain or path relative to your application URL
+2. Post-Password Redirect URL
+3. Post-Profile Redirect URL
+
+To overwrite the Global Redirect URL set the Default Redirect URL to a fully
```
aeneasr, line 61: I find this a bit confusing, what's the difference between "Global Redirect URL" and "Default Redirect URL"?

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,125 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
+qualified domain name
+([FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) - for example
+`https://domain.example/login`).  
+To replace the Global Redirect URL path set the Default Redirect URL to a
+path. For example when setting it to `/login`,
+`https://domain.example/global/path` becomes `https://domain.example/login`.
```
aeneasr, line 67: I don't quite understand what this means? Are you trying to explain that you can set both an absolute (`https://www.example.org/some/path`) and relative (`/some/path`) URL?

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,125 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
+The Default Redirect URL is used when Ory isn't sure where to redirect. It’s a
+good idea for this to be your default app URL, for example `/` or the dashboard.
+Use sub-conditions to control where users are redirected after registration,
+login, and other self-service flows.
+
+On sub-conditions such as After Password or After Profile an FQDN or a path
+can be set.  
+Set the Post- redirect to a FQDN to replace the closest relative, starting
+with the Login/Registration Default Redirect URL and then the
+Global Redirect URL.
```
aeneasr, line 78: I think we should call just call this `Default Redirect URL` and not introduce `Global`, I think they refer to the same thing?

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,125 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
+![Browser Redirects URL Error Message](/images/browser-redirects/browser-redirects-error-messages-1.png)
 
 ![Browser Redirects URL Error Message Flows](/images/browser-redirects/browser-redirects-error-messages-flows.png)
 
 ### Domain Denied
 
-URLs matching any of the denied URLs, such as the Ory domain will automatically
-be denied. This deny list can be extended in future to prevent any abuse on the
-Ory Cloud.
+It's not possible to set the `ory` domain as redirect URL.
```
aeneasr, line 153:
```suggestion
It's not possible to set the any Ory-owned domain as redirect URL.
```

aeneasr (changes requested):

docs/concepts/redirects.mdx
vinckr: I kind of went with the wording in the original doc - which follows the names in the UI.

However I think it makes sense to omit the Global and call it the Default Redirect URL.

Screenshot 2022-03-18 at 17 13 19

"Global Redirect URL" would be the "global default redirect URL" in the very top 

"Default Redirect URL" means the "Post- Default Redirect URLs", so for example "Post-Login Default Redirect URL".

That is the source of confusion. It is probably sensible to give the Post-X flows a common name to avoid this confusion? Maybe "Default Redirect URLs for Self-Service Flows"?
and then the Default Redirect URL is the global one.

docs/concepts/redirects.mdx
vinckr: See here : https://github.com/ory/docs/pull/668/files#r830157781

 I see to clarify this a bit and see how it works without the Global.

aeneasr: Oh I see, so the key in Ory Kratos is always `default_browser_return_url`, and then we have different prefixes for it. I think this is a bit confusing when reading it in the docs. It might be a good idea to present this in a diagram or something - even a YAML example would help.

Maybe this document helps also?

https://www.ory.sh/docs/kratos/concepts/browser-redirect-flow-completion

docs/concepts/redirects.mdx
aeneasr: How about "Redirect target after successful <login|registration|...>" and "Redirect target after successful <login|registration|...> with <password|social|...>"?

vinckr:
I removed the "Global" Redirects term, so its consistently Default Redirects URL and the others are called Post-Login Redirect URL and After Password Redirect URL etc.

I also added the .yml code example with some comments. 

I wanted to have a bigger .yml that works both for OSS and Cloud documentation, but we need to be able to display only specific lines (so we dont have 10 .yml files for this). Maybe it is also a bit overkill.

tomekpapiernik (approved):

docs/concepts/redirects.mdx
```diff
@@ -40,105 +38,148 @@ will fail, however, for https://anotherapp.com will succeed on any path.
...
-can be set. In the case of an FQDN it will replace the closest relative starting
-with the Login/Registration `Default Redirect URL` and then the
-`Global Redirect URL`. With a path it will use the closest relative as the base
-URL as described above with the `Default Redirect URL`.
+1. Default Redirect URL domain or path relative to your application URL
+2. Post-Password Redirect URL
+3. Post-OIDC Redirect URL
+
+```yaml file="kratos.config.yml"
+selfservice:
```
aeneasr, line 56: As a follow up we should show how to set this using tabs:

1. Self-Hosted Kratos (this config file here)
2. Ory Console (screenshot)
3. Ory CLI (`ory patch project .. --replace /.../...`)

I'm ok not doing this in this PR but can you please create a follow up issue and tackle this soon? Because this section is otherwise not very helpful for cloud customers. Please link the issue here

aeneasr (approved):
Nice!

PR #732, opened 2022-04-01
fix: general contributing guide

vinckr:
- fix: general contributing doc
- fix: remove other contributing docs

Adds the contributing guidelines to the guide in /ecosystem.
Removes the individual guides for kratos/hydra/oathkeeper (for some reason keto didnt have one)

I will open another PR in /web for the redirects, or do we set them in https://github.com/ory/docs/blob/master/vercel.json ?


docs/ecosystem/contributing.md
```diff
@@ -1,15 +1,300 @@
...
+    - [Responses](#responses)
+- [IDE Tips](#ide-tips)
+  - [Goland](#goland)
+    - [Goland Tests](#goland-tests)
+  - [Jetbrains](#jetbrains)
+    - [Debugging Tests](#debugging-tests)
+  - [VS Code](#vs-code)
+    - [VS Code Tests](#vs-code-tests)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```
tomekpapiernik, line 58: It doesn't make much sense to add a ToC when we want this document to be consumed through the website. Docusaurus generates a ToC from headings automatically and displays it as right-hand navigation. Please remove the in-doc toc you added. 

image

src/sidebar.js
```diff
@@ -134,7 +134,6 @@ module.exports = {
             'kratos/quickstart',
             'kratos/install',
             'kratos/configuring',
-            'kratos/contributing'
```
tomekpapiernik, line 137: I think you should add redirects for the pages that get removed in the https://github.com/ory/docs/blob/master/vercel.json file. That's what I was asked to do when I removed/moved docs around ;)

tomekpapiernik, line 137: I think you should add redirects for the pages that get removed in the https://github.com/ory/docs/blob/master/vercel.json file. That's what I was asked to do when I removed/moved docs around ;)

tomekpapiernik (changes requested):

docs/ecosystem/contributing.md
vinckr: thanks, yea that does not make any sense 🙈

src/sidebar.js
vinckr: I added them here: 
https://github.com/ory/web/pull/613

But I was also not sure if I should use the vercel.json or the _redirects in ory/web. 
I guess it does not matter? 
(other than us getting confused because there is so many redirects and two files).
 
Maybe we should do all /docs redirects in the vercel.json and the rest in ory/web?

Happy to add it to vercel.json, but lets come up with a rule so its clear where to add which redirect 🙏

tomekpapiernik: Oh man, I'm not sure what's the convention. If it were up to me I'd stick to handling redirects for the documentation in this repo and limit the other file to website-related redirects only. I'm not sure why everything's structured the way it is now.

vinckr: Ok I will add the redirects to the vercel.json here and then we handle docs redirects there going forward. 
We should then also move the other redirects for docs here.

vinckr:
Tests are failing but unrelated to this PR :/

tomekpapiernik (approved):

PR #754, opened 2022-04-12
docs: refactor and update docs for proxy and tunnel

tomekpapiernik:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

Updated and refactored docs around the Proxy part of the CLI. Added some info on Tunnel.


docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+If you want to adjust the entry point URL, pass the desired address along with the application URL, 
+for example: 
+
+```shell
+ory proxy http://localhost:3000 http://localhost:3001
+```
+
+### Usage in Production
+
+You can use Ory Proxy in a production setup when you are self-hosting Ory components.
```
vinckr, line 183: what do you mean by components? 
I thought you just self-host the user interface / frontend? 
I think ory proxy only works with ory cloud.

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+```html
+<a href="/self-service/login/browser">Log in</a>
+```
+
+  </TabItem>
+</Tabs>
+````
+
+You can also append a `return_to` query parameter to the URL. This redirects the user
+to the set URL when they complete the self-service flow.
```
vinckr, line 347:
```suggestion
You can also append a `return_to=<your-url>` query parameter to the URL. This redirects the user
to the set URL when they complete the self-service flow.
```

maybe add this so its clear what you mean by "set URL"?

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+:::
+
+## Ory Proxy 
+
+Ory Proxy is a reverse proxy deployed in front of your application. This allows the 
+Ory endpoints to be mirrored on the same domain as the app you're running. 
+Additionally, the Proxy converts sessions into JSON Web Tokens (JWTs) and ensures that cookies 
+and URLs are properly configured.
+
+:::info
```
gen1us2k, line 35: I vote for some diagrams to show how this works to make understanding easier because using ory proxy without additional context is quite hard. I can create this one for you to explain how it works.
Also, it would be great to add the motivation part why do I need to use ory proxy, for instance

- to prevent CORS issues
- to have integration with the local environment and ory cloud project

Also, I think that we need to have some examples of using ory proxy. Here're a couple of guides that can help you

- https://www.ory.sh/docs/guides/proxy
- https://www.ory.sh/docs/guides/proxy#ory-proxy-for-local-development

tomekpapiernik: Thanks, Andrew. If you want to whip up a diagram then I'm happy to include it :) 

When it comes to the motivation, have a look at lines 30-44 - I think this is covered. I can add a mention of the CORS aspect.

> Also, I think that we need to have some examples of using ory proxy.

This document gets removed and replaced by this one. Some of the content is moved from that document to the new one with some edits. 

When it comes to examples, the proxy is featured in almost all of the "protect page with login" examples and I don't really feel that this is a place to whip up yet another example - you'd have to describe a setup with an app, etc. This section is supposed to be all about CLI, no clutter ;)

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
+---
+id: proxy-and-tunnel
+title: Ory Proxy and Ory Tunnel
+---
+
+The CLI contains two important tools for local development and self-hosted instances
+of the Ory Identity Service (Kratos):
```
Benehiko, line 7: It doesn't really have anything to do with self-hosted instances, but it can be used for it. On self-hosted instances you could just deploy your own Kratos locally as well for testing, so it will be a rare occurrence.

The proxy/tunnel is there to help you with your application development locally and can also be there for you in production when you do not have your domain added to Ory. The problem this is solving is to circumvent the browser CORS policy. Remember if you use a tool like `curl` you can communicate fine with the Ory APIs, but if you are doing it the browser it will complain because your current domain (e.g. localhost:4000 or www.example.com) doesn't match up with the ory domain (e.g. your-slug.projects.oryapis.com). 

So the proxy/tunnel is now the new `curl` doing the requests outside of the browser's scope and your browser is doing requests to the proxy/tunnel directly which is also "hosting" your application for you on the same domain (e.g. localhost:4000 or www.example.com)

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+navigation panel, and copy the URL from the **SDK Configuration** section.
+
+:::
+
+## Ory Proxy 
+
+Ory Proxy is a reverse proxy deployed in front of your application. This allows the 
+Ory endpoints to be mirrored on the same domain as the app you're running. 
+Additionally, the Proxy converts sessions into JSON Web Tokens (JWTs) and ensures that cookies 
+and URLs are properly configured.
```
Benehiko, line 33: Yes, but as @gen1us2k said, we need to be a  bit more informative here than just the general "it helps you". 

See my comment above ^

The proxy rewrites the cookies to match the domain your application is currently on (either the proxy domain of localhost:4000 or example.com). So instead of the cookie being set to `project-slug.projects.oryapis.com` it is now correctly set to the domain you own. (with cname this isn't necessary - since their Ory project is served under the domain name they own, but for local development they will still need o use the proxy/tunnel)

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+
+:::info
+
+Using the Proxy is an alternative to custom domains (CNAME), and a useful tool when developing locally. 
+Even though you can use the Proxy in production, we recommend using our Custom Domain feature where possible.
```
Benehiko, line 38: Add something here that deploying to production is more seamless when using CNAME - with the ory proxy you need to think about additional setup and configuration

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+
+:::
+
+When an application runs on `http://localhost:3000`, Ory must be available 
+on the same domain, for example at `http://localhost:3001`. Thanks to such setup, 
+the system can manage Anti-CSRF Cookies and Ory Session Cookies.
```
Benehiko, line 44: this doesn't make sense. localhost:3000 and localhost:3001 are different origins :D 

I think you should also mention origins and explain what they are. CORS errors are the reason for different origins.
https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+GET /some-path
+Host: localhost:4000
+Cookie: ory_session_jollyproskuriakovaxe98qw5t8g=MTYyNzU1OTgyNHxEdi1CQkFFQ180SUFBUkFCRUFBQVJfLUNBQUVHYzNSeWFXNW5EQThBRFhObGMzTnBiMjVmZEc5clpXNEdjM1J5YVc1bkRDSUFJR3RGU1d4dlUwOXVSR2w1UjJONmFVRlhaWEIxWVhCVlNHWlZOVTQxWWtGMnwhbFZh8BCCQ3tMemDczrB9-epefXl1E7whiChUt62LuA==
+User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.164 Safari/537.36
+...
+```
+
+Ory Proxy resolves the session and converts it to a JSON Web Token. The token is
+included in the `Authorization` HTTP header of the request made to the
+"upstream" (the URL protected by the Ory Proxy).
```
Benehiko, line 65: add here - "your application"

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+
+For example, if you are running a NodeJS server on port `3000` and you expose web
+traffic on port `8080`: 
+
+```shell
+ory proxy --port 8080 http://127.0.0.1:3000 https://your-domain.com &
+
+# This should be your server command
+node your-entrypoint.js
+```
```
Benehiko, line 203: Questionable - always have something like NGINX or Traefik in front of your deployment. The proxy should be exposed directly, this makes it difficult to do maintenance and load balancing. The proxy does not offer the same functionality as NGINX / Traefik etc. 

You can still run this command as is since this will rewrite all cookies etc to match `https://your-domain.com`. There is also a flag that you can set for which domain the cookies should be set on if you are running this on a subdomain `https://sub.your-domain.com`. 

```shell
ory proxy --cookie-domain example.org \
		http://127.0.0.1:3000 \
		https://ory.example.org
```

Benehiko: You then need to point NGINX / Traefik to your proxy url `http://localhost:4000` unless you changed the port with `--port 3001`.

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,376 @@
...
+## URL Structure
+
+:::tip
+Ory Proxy adds `/.ory` prefix when mirroring the Ory Cloud APIs and UIs. 
+:::
+
+For example, when using the Ory Proxy, calling 
+`https://<proxy-host>/.ory/ui/login` is the same as calling `https://<your-project-slug>.projects.oryapis.com/ui/login`
+directly.
+
```
Benehiko, line 288: why are we talking about the Ory proxy again here?

Benehiko (changes requested):

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: This is something I ported over from the old doc: 
https://www.ory.sh/docs/guides/proxy#url-structure

What's wrong with this section?

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: I understand what you mean with the need for something like NGINX, but this is just a simple example of how you could use the proxy if you really wanted to. Wouldn't it work if you looked at it like that?

docs/guides/cli/20_proxy-and-tunnel.mdx
Benehiko: since we already spoke about the proxy in the previous section this might be a bit confusing here? Maybe move this up to the `proxy` section?

docs/guides/cli/20_proxy-and-tunnel.mdx
Benehiko: I think just note here that you would run your ingress proxy in front of the ory proxy (such as NGINX etc)

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: Still, if you look at the markdown table, there's a version that shows how the self-service URLs work in Tunnel (no prefix) and that's why I made it a separate section - it applies to both the Proxy and the Tunnel. But I guess it doesn't make much sense to include that? I mean, the tunnel doesn't alter the default self-service URLs?

Benehiko: yeah the tunnel doesn't alter it. i think you can though through the a flag

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: done

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: done

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: Okay, I moved this section under the Proxy and removed the switches to toggle between the proxy and the tunnel since the tunnel doesn't affect the regular self-service URLs.

Benehiko (changes requested):
Looking a lot better! just some minor stuff still and then we can merge it :)

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -0,0 +1,343 @@
...
+instead of `<your-project-slug>.projects.oryapis.com`. 
+
+This behavior is necessary to avoid issues with the browser [CORS policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).
+
+Additionally, the Proxy converts sessions into JSON Web Tokens (JWTs) and ensures that cookies 
+and URLs are properly configured.
+
+:::info
+
+Using the Proxy is an alternative to custom domains (CNAME), and a useful tool when developing locally.
```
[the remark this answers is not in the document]
tomekpapiernik: sure!

Benehiko (approved):
Nice! LGTM :)

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -323,13 +321,17 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
...
 
-The Ory Tunnel has the same functionality and behaves the same way 
-as the Proxy with these exceptions:  
+The main role of the Ory Tunnel is exposing the Ory APIs under the same top-level domain 
+as your application to ensure that there are no CORS issues.
+
+The Tunnel shares most of its code with the Ory Proxy, and from the user's point of view
+it works in a similar way. There are, however, several important differences between these tools: 
 
+  ❌ Ory Tunnel doesn't handle traffic routing
```
Benehiko, line 330: what does that mean?

docs/guides/cli/20_proxy-and-tunnel.mdx
```diff
@@ -323,13 +321,17 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
...
+  ❌ Ory Tunnel doesn't mirror Ory URLs and as such doesn't add the `.ory` prefix to any of the Ory URLs
```
Benehiko, line 331: ah so it doesn't mirror the UI urls? or only the APIs? not sure - a bit confused by this sentence

Benehiko (changes requested):
The differences between `tunnel` and `proxy` don't seem clear by the additional comparisons.

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: The tunnel doesn't act as an entry point to the app the way I understand it so you'd call the app directly instead of the proxy. Then the app gets to the Ory endpoints through the tunnel. With the proxy, the traffic hits the proxy and it then "manages" the request, takes it to the app and then to the endpoints. That's what I got after talking to you and PM. Does that make sense?

tomekpapiernik: From `--help`:

```
It is similar to the `ory proxy` command, but it does not require to
route all your application's traffic through it!
```

docs/guides/cli/20_proxy-and-tunnel.mdx
tomekpapiernik: Not sure how to answer that, need support from someone more dev-savvy :<

piotrmsc: it does not mirror but tunnel the API's - there's no `.ory/` prefix that you need to add to path when calling an endpoint. 
UI is available under `localhost:4000/ui/welcome` as well you can initiate login flow by calling `localhost:4000/self-service/login/browser` :)

piotrmsc (approved):

docs/guides/cli/20_proxy-and-tunnel.mdx
Benehiko: i see

docs/guides/cli/20_proxy-and-tunnel.mdx
Benehiko: But the traffic still goes through the tunnel to your app?

Benehiko: I understand that it does not route the traffic through the `.ory` prefix on the URL and forwards the requests to the Ory apis on their natural paths.

But the part about `Ory tunnel doesn't handle traffic routing` sounds wrong since the tunnel does indeed handle traffic to the Ory Apis (the request goes through the tunnel to get there).

From the perspective of the App not being behind the tunnel i can understand, but then why would you need to provide your App URL to the tunnel?

Benehiko: Ah okay i just reread the documentation from the cli. 
I see it can be run separate from your application. 
I think you need to reword this sentence though


❌ Ory Tunnel doesn't handle traffic routing to your application, but instead to Ory APIs.

PR #775, opened 2022-04-21
chore: disable vale linter

vinckr:
- remove the vale linter workflow as discussed with @tomekpapiernik 

We might add a style linter down the line with individual styles, as of now the write-good (or other style collections) are too broad to be enforced (and not useful if not enforced)


tomekpapiernik (approved):
This will make docu reviews easier on the eyes, thanks :) I hope we'll be able to implement a basic spelling check in place of this set of rules soon. Will be more useful IMO

aeneasr:
Yes, can we add at least some basic spell checking again? I found it quite useful. It's also good for brand control (eg Ory and not ORY)

vinckr:
Markdownlint is still active: 
https://github.com/ory/docs/blob/master/docs/markdownlint.yml
This also includes `proper-names`, where things like Ory/ORY get checked. 

So what we removed here was just the [write-good rules](https://github.com/errata-ai/write-good/tree/master/write-good).

You can always run vale locally if you like the write-good rules. I have a vale.ini in my home folder and just do `vale documentpath` before pushing. 

I added a draft PR for a proper spellchecker (something we did not have before): 
 https://github.com/ory/docs/pull/778

PR #783, opened 2022-04-26
docs: update ory cli description for protect page with login guides

tomekpapiernik:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

Updated the reusable Ory CLI snippet used across the `protect page with login` examples to explain the role of the proxy in a simpler way and to refer to the recently added CLI proxy/tunnel documentation.

### DIFF
--- a/docs/guides/_common/ory-cli.mdx
+++ b/docs/guides/_common/ory-cli.mdx
@@ -7,7 +7,7 @@ domain as the application you're running and rewrites cookies to match the domai
 As a result, the origin of the cookies is set correctly to the domain you run the app on instead of <br /> 
 `<your-project-slug>.projects.oryapis.com.`
 
-This behavior is necessary to avoid issues with the browser [CORS policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).
+This behavior is necessary to avoid issues with the browser [CORS policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
 
 By using the Proxy, you can easily connect the application you're developing locally to Ory Cloud and consume
 the APIs without the hassle of additional configuration and without the need to self-host any of the Ory Components. 
--- a/docs/guides/_common/ory-cli.mdx
+++ b/docs/guides/_common/ory-cli.mdx
@@ -1,18 +1,21 @@
-The Ory CLI includes useful functionality to manage your Ory Cloud Project. But
-that is not why we require it in this guide!
+Ory CLI provides a convenient way to configure and manage Ory Cloud projects. Additionally, the CLI 
+contains Ory Proxy - a reverse proxy that rewrites cookies to match the domain your application is currently on. 
 
-Ory's philosophy is to make hard things easy for you. For this reason, Ory has
-deployed measures against all
-[OWASP Top 10](https://owasp.org/www-project-top-ten/) and implements the
-[OWASP Authentication Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html)
-along other mechanisms.
+Ory Proxy is a reverse proxy deployed in front of your application. The Proxy mirrors Ory endpoints on the same 
+domain as the application you're running and rewrites cookies to match the domain your application is currently on.
 
-Therefore, Ory manages Anti-CSRF Cookies as well as Ory Session Cookies for you.
-That however requires that Ory and your application run on the same domain!
+As a result, the origin of the cookies is set correctly to the domain you run the app on instead of <br /> 
+`<your-project-slug>.projects.oryapis.com.`
 
-If your application runs on `http://localhost:3000` then Ory needs to be
-available on the hostname `localhost` as well (e.g. `http://localhost:3001`).
-That is why we need the Ory CLI, because it has a proxy included which mirrors
-Ory's API endpoints on the domain of your application.
+This behavior is necessary to avoid issues with the browser [CORS policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).
+
+By using the Proxy, you can easily connect the application you're developing locally to Ory Cloud and consume
+the APIs without the hassle of additional configuration and without the need to self-host any of the Ory Components. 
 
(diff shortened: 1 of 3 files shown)


docs/guides/_common/ory-cli.mdx
```diff
@@ -1,18 +1,21 @@
-The Ory CLI includes useful functionality to manage your Ory Cloud Project. But
-that is not why we require it in this guide!
+Ory CLI provides a convenient way to configure and manage Ory Cloud projects. Additionally, the CLI 
+contains Ory Proxy - a reverse proxy that rewrites cookies to match the domain your application is currently on. 
 
-Ory's philosophy is to make hard things easy for you. For this reason, Ory has
-deployed measures against all
-[OWASP Top 10](https://owasp.org/www-project-top-ten/) and implements the
```
aeneasr, line 6: In the past, people have been quite receptive to claims that we take care of many security concerns (such as the ones in OWASP). For example, Anti-CSRF measures are annoying when setting it up, but life saving when protecting against CSRF attacks. It doesn't have to be in the first paragraph, but I think we should at least mention that "Ory keeps you safe", and not just "Ory sets the domain right" :)

aeneasr (changes requested):

docs/guides/_common/ory-cli.mdx
```diff
@@ -1,18 +1,21 @@
-The Ory CLI includes useful functionality to manage your Ory Cloud Project. But
-that is not why we require it in this guide!
+Ory CLI provides a convenient way to configure and manage Ory Cloud projects. Additionally, the CLI 
+contains Ory Proxy - a reverse proxy that rewrites cookies to match the domain your application is currently on.
```
piotrmsc, line 2: can we now also make the CLI description unified in places such as :  https://docs-gzre1hl01-ory.vercel.app/docs/guides/cli/installation ?

docs/guides/_common/ory-cli.mdx
tomekpapiernik: I can do that if you REALLY want to see it mentioned here, but consider this - this is a snippet used on a very specific set of documents only. These documents are examples that show you very simple apps consuming Ory. The question here is - why do I, the user, even need the CLI to work with these examples? 
The way I structured the description aims to answer this question and bring attention to two things that are the most important in that context: 
- the fact that the proxy "connects you to the cloud so you don't need to install anything else on your machine to see the setup working" 
- the fact that the endpoints are mirrored on the same domain and cookie origin is setup correctly thanks to that

That's why I cut the "Ory keeps you safe" elements :)

docs/guides/_common/ory-cli.mdx
```diff
@@ -1,18 +1,21 @@
...
-Therefore, Ory manages Anti-CSRF Cookies as well as Ory Session Cookies for you.
-That however requires that Ory and your application run on the same domain!
+As a result, the origin of the cookies is set correctly to the domain you run the app on instead of <br /> 
+`<your-project-slug>.projects.oryapis.com.`
 
-If your application runs on `http://localhost:3000` then Ory needs to be
-available on the hostname `localhost` as well (e.g. `http://localhost:3001`).
-That is why we need the Ory CLI, because it has a proxy included which mirrors
-Ory's API endpoints on the domain of your application.
+This behavior is necessary to avoid issues with the browser [CORS policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).
```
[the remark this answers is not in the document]
tomekpapiernik: Updated the link to

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

docs/guides/_common/ory-cli.mdx
tomekpapiernik: I wouldn't do that. The CLI desc. in this snippet is very condensed and tailored to fit this specific context. The dedicated section has more "breathing room" so the descriptions look different.

docs/guides/_common/ory-cli.mdx
aeneasr: That makes sense, maybe we can move it to a separate document then? For example `Security Architecture` or something like that? And have an overview of the different security relevant topics for nerds? Similar to: https://www.ory.sh/docs/hydra/security-architecture

tomekpapiernik: Yes, I'd rather do it this way. :)

docs/guides/_common/ory-cli.mdx
```diff
@@ -1,18 +1,21 @@
...
-[OWASP Top 10](https://owasp.org/www-project-top-ten/) and implements the
-[OWASP Authentication Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html)
-along other mechanisms.
+Ory Proxy is a reverse proxy deployed in front of your application. The Proxy mirrors Ory endpoints on the same 
+domain as the application you're running and rewrites cookies to match the domain your application is currently on.
 
-Therefore, Ory manages Anti-CSRF Cookies as well as Ory Session Cookies for you.
-That however requires that Ory and your application run on the same domain!
+As a result, the origin of the cookies is set correctly to the domain you run the app on instead of <br /> 
+`<your-project-slug>.projects.oryapis.com.`
```
aeneasr, line 8: You can add newlines using two white spaces at the end:

```
instead of  
```

however, it usually does not look very great visually to have a singular code box, so it is usually better to have a code box for this (as it allows you to copy the code also):

````suggestion
As a result, the origin of the cookies is set correctly to the domain you run the app on instead of <br /> 

```
<your-project-slug>.projects.oryapis.com.
```
````

aeneasr (commented):
Almost LGTM :) Only one nitpick

docs/guides/_common/ory-cli.mdx
tomekpapiernik: Oh, I did that bc for some reason docusaurus broke the address into two lines which made it look weird. 
Makes sense to put it into a code block, I'll commit your suggestion.

aeneasr: 👍

PR #807, opened 2022-05-12
fix: update account linking-related content

tomekpapiernik:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/docs/guides/social-signin/overview.mdx
+++ b/docs/guides/social-signin/overview.mdx
@@ -40,9 +40,9 @@ Sessions. The tokens are used in the following way:
 
 ## Preventing Account Linking
 
-If you enable multiple registration methods, users may be redirected to a social login provider to link their existing accounts
-after email verification. To disable this behavior, add the `session` hook after one of the registration methods (for example
-`password`):
+If you enable multiple registration methods, users can link their existing accounts
+to social sign-in provider accounts after email verification. To disable this behavior, 
+add the `session` hook after one of the registration methods (for example `password`):
 
 ```yaml
 selfservice:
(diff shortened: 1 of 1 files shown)


docs/guides/social-signin/overview.mdx
```diff
@@ -65,16 +64,7 @@ selfservice:
...
-account that has the same identifier as an existing account, the user is asked
-to link the accounts.
-
-There is no way to have two separate accounts (one created "manually", one
-through a social sign-in) for a single identifier.
-
-This is the default behavior of the system.
-
-While convenient for users, this behavior creates an attack vector that can
+While convenient for users, linking accounts creates an attack vector that can
```
tomekpapiernik, line 67: Hey @aeneasr, could you please help me out with this section? 
Since the linking operation has to be engaged manually by the user in their account settings, 
I don't see how a malicious actor could take over a "manually created account" in Ory by taking over the user's 
social sign-in account (e.g. GH). 

The way I understand it, such a takeover would be possible provided account linking is an automated operation - but it's not. 

With that said, with my current state of understanding of this topic, the `Dangers of Account Linking` section doesn't really make sense and should be deleted asap. 

Am I missing something there? If so, would you mind explaining?

aeneasr: You're right, how Ory solves this currently, this is not a risk:

> While convenient for users, linking accounts creates an attack vector that can
> allow malicious actors to take over user accounts.

I think this section was originally about the login process itself - here we must not automatically link two social profiles together. But later it got mixed in the terminology with how we implemented social account linking securely, namely by going to your settings and linking the provider there.

So I agree, this section should focus on:

1. It is not possible to link accounts automatically durign login and registration due to this attack vector
2. Here is how we still support account linking: By going to settings and linking there

Makes sense?

tomekpapiernik:
@aeneasr What do you think, how does it look now?

aeneasr (approved):

PR #813, opened 2022-05-16
feat: low-latency session validation guide

gen1us2k:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).


code-examples/low-latency-sessions/main.go
```diff
@@ -0,0 +1,121 @@
...
+func NewMiddleware() *kratosMiddleware {
+	configuration := client.NewConfiguration()
+	configuration.Servers = []client.ServerConfiguration{
+		{
+			URL: "https://playground.projects.oryapis.com",
+		},
+	}
+	return &kratosMiddleware{
+		client: client.NewAPIClient(configuration),
+		redis: redis.NewClient(&redis.Options{
```
aeneasr, line 31: I think this is really cool to use redis! An alternative, which would probably be a little bit faster, would be to use something like:

- https://github.com/hashicorp/golang-lru
- https://dgraph.io/blog/post/caching-in-go/

That would make the tutorial a bit easier because you don't need to set up redis.

However, redis is of course better when in production. What do you think?

gen1us2k: golang-lru, goleveldb and other in-memory embeddable solutions are great to have in the application. However, for production usage, people use Redis or Memcached. I'll add an example for in-memory caches but I'll add a couple of notes, like this does not scale well and you need to use redis or something similar in production.

aeneasr: That's a really good idea, thanks :)

gen1us2k: I added an example of using ristretto cache. golang-lru does not support TTL out of the box.

aeneasr: Even better :)

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+Ory Cloud is an external dependency, and latency to validate a session may vary. Here's the example
+
+```bash
+curl -o /dev/null\
+     -w %{time_connect}:%{time_starttransfer}:%{time_total} \
+     -s https://playground.projects.oryapis.com/sessions/whoami
+     -H 'ory_session_playground=MTY1MjcyMjk5OXxBenA5MVdvaXpPZ242QlZtOVNCZ2JuRWt4NVB4WFFIVGdLcmFIZl9BN0VTOHZ3VHlJbVdVenpaZl96VWtWS1p4clI3bFRkYnFBeDZwdFZGU3RvSmxyY0JpUGJLNGQwLUlfWnNCMlFlMS1CbDlBdEtjX0xoLWxmSUNmWUJWU24xWTBoNlBGM0laSHc9PXyOu8_5YP6pFxhuNLWD3fmzxdo8JWo27iy0NkF2HQ7fcA=='
+
+0.275457:0.778495:0.778623
```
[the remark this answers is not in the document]
gen1us2k: It depends on the region and last-mile bandwidth. I'll make some adjustments.

piotrmsc:
hey @gen1us2k  is it ready for another round of review?

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+1. Using embedded in-memory solutions can give one low latency. However, an application stores state can prevent horizontally scaling it.
+1. Key-value storages add network latency, but an application does not store any state so that one can scale it without any issues.
+
+:::warning
+
+Be careful with the TTL for cached sessions, and do not forget to remove the session once it expires or a user logs out. Users can still have access because of cached data, leading to a potential security breach.
+
+:::
+
+Let us take a look at following examples. We have session middleware that validates sessions and caches data. The logic is simple
```
tomekpapiernik, line 60: > Let us take a look at following examples

Are these examples? I don't think so. 

> The logic is simple

Add punctuation.

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+1. Try to get cached session from Redis;
+1. If a session is valid, return it and do not request Ory Cloud;
+1. If data does not persist, validate a session against Ory Cloud and cache it.
```
tomekpapiernik, line 64: https://docs.microsoft.com/en-us/style-guide/scannable-content/lists#punctuation

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+## Network architecture
+
+1. `8080` port is used by the REST API
+1. `4000` port is used by [`ory proxy`](https://ory.sh/docs/guides/cli/proxy-and-tunnel) that manages cookies for local domain and proxies traffic to Ory Cloud
```
tomekpapiernik, line 69: As someone who isn't really familiar with the topic and your train of thought, I don't understand what's this section doing here. What's the purpose of talking about the proxy?

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+```mdx-code-block title="main.go"
+import index from '!!raw-loader!../../../code-examples/low-latency-sessions/main.go'
+
+<CodeBlock language="go">{index}</CodeBlock>
+```
+The strategy above works for most cases when one has a high-throughput system.
+
+:::warning
+Caching can worsen the situation in the low-throughput systems because of [RTT](https://en.wikipedia.org/wiki/Round-trip_delay)
```
tomekpapiernik, line 100: When using abbreviated names of concepts/technologies, try to present them in full length first. Here you use RTT - use `round-trip time (RTT)` instead or skip the abbreviation altogether.

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+A performance difference. Open `http://localhost:8080/ping`
+
+```
+# Without caching the response time is 1.5s
+[GIN] 2022/05/17 - 18:27:58 | 200 |  1.501481333s |             ::1 | GET      "/ping"
+
+# With caching enabled the response time is 5 ms
+[GIN] 2022/05/17 - 18:28:42 | 200 |    5.309667ms |             ::1 | GET      "/ping"
+```
```
tomekpapiernik, line 112: Where's this difference coming from? What are you comparing?

docs/guides/session-management/low-latency-session-management.mdx
```diff
@@ -0,0 +1,113 @@
+---
```
tomekpapiernik, line 1: General remark - heading structure is flat - H2 all across the doc. Maybe you should use H3s too to organize the content better?

code-examples/low-latency-sessions/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	// For local development, it's requires to use
+	// ory proxy to make cookies work fine.
+	// Replace it with http://projectId.projects.oryapis.com
+	// for production usage.
+	sdkURL = "http://localhost:4000/.ory"
+	// For local development, it's requires to use
+	// ory proxy to make cookies work fine.
+	// Replace it with http://projectId.projects.oryapis.com/ui/login
```
tomekpapiernik, line 25: instead of `projectId` please use `<project_slug>` - this form is used in many places in our docs already

code-examples/low-latency-sessions/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/go-redis/redis/v8"
+
+	client "github.com/ory/kratos-client-go"
+)
+
+const (
+	// For local development, it's requires to use
+	// ory proxy to make cookies work fine.
```
tomekpapiernik, line 19:
```suggestion
	// When developing locally, use Ory Proxy
	// to have the right origin set in HTTP cookies.
```

Maybe instead of the proxy we could promote the Tunnel here?

code-examples/low-latency-sessions/embedded/main.go
```diff
@@ -0,0 +1,137 @@
...
+			c.Redirect(http.StatusFound, "http://your_endpoint")
+			return
+		}
+		c.Next()
+	}
+}
+func (k *kratosMiddleware) cacheSession(key string, sess *client.Session) error {
+	// Calculate TTL. Do not cache sessions without TTL.
+	//
+	// Cached session should evict on sess.ExpiresAt
```
tomekpapiernik, line 75: evict? 
You mean the session is removed from the cache?

tomekpapiernik (changes requested):

code-examples/low-latency-sessions/embedded/main.go
gen1us2k: Yep

docs/guides/session-management/low-latency-session-management.mdx
gen1us2k: It's coming from using cache. See comments #Without caching and #with caching

docs/guides/session-management/low-latency-session-management.mdx
gen1us2k: Added H3 headings

code-examples/low-latency-sessions/main.go
gen1us2k: Done

code-examples/low-latency-sessions/embedded/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}, nil
+}
+func (k *kratosMiddleware) Session() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		session, err := k.validateSession(c.Request)
+		if err != nil {
+			c.Redirect(http.StatusFound, loginURL)
+			return
+		}
+		if !*session.Active {
```
piotrmsc, line 65: If session is not active shouldn't we redirect to ory in order to authenticate?

code-examples/low-latency-sessions/embedded/main.go
```diff
@@ -0,0 +1,137 @@
...
+	r := gin.Default()
+	k, err := NewMiddleware()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	r.Use(k.Session())
+	r.GET("/ping", func(c *gin.Context) {
+		c.JSON(200, gin.H{
+			"message": "pong",
```
piotrmsc, line 133: There's a tiny bug here :) 

```
[GIN-debug] [WARNING] Headers were already written. Wanted to override status code 301 with 200
```

gen1us2k:
Hey @piotrmsc 

> In the tutorial we should describe what URL users should set and what URLs should use. It's not described what endpoint user should call (/ or /ping etc.). We should describe also that we need to override certain params like http://your_endpoint -> here would be nice to support environment variables for example.

Now it's redirect users to the Ory login page

> To allow user experience better the performance improvement of caching I would propose to have 2 separate endpoints where one : does not implement caching at all b) second uses a cache and if a user make a request to first one and then to second one he can easily in the terminal logs see the timings :)

Agreed. I made small changes to the source code of both examples. 

> In the code directory you have a sub directory called embedded, looking by code it focuses on in-memory cache implementation but this is not mentioned in the tutorial. It's not clear whether I should use it or not and how :)

It is mentioned and it goes first because it's easier to use and see the difference.

vinckr (commented):
I cant comment on the code, but the doc looks good 👍

PR #822, opened 2022-05-23
docs: add security model doc

aeneasr:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).


docs/guides/local-development.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+### Social Sign In
+
+Setting up Social Sign In on `localhost` is not yet officially supported. We are working towards a solution that allows us to
+easily set up Social Sign In when developing locally.
+
+### WebAuthn
+
+Similar to Social Sign In, WebAuthn is not yet supported on `localhost`. The reason is the explicit origin matching in WebAuthn
+which is a measure against phishing. We are working towards a solution to make WebAuthn available on localhost as well!
```
tomekpapiernik, line 57:
```suggestion
### Social Sign In

Setting up Social Sign In on `localhost` is not yet officially supported. We are working towards a solution that allows us to
easily set up Social Sign In when developing locally.

### WebAuthn

Similar to Social Sign In, WebAuthn is not yet supported on `localhost`. The reason is the explicit origin matching in WebAuthn
which is a measure against phishing. We are working towards a solution to make WebAuthn available on localhost as well!
```

I think that it'd be a good idea to add these two notes to the respective topics in our documentation, wdyt? 
It could save time for people who learn that we have these two features available and decide to test them right away with Kratos running locally.

tomekpapiernik: Hm, I'm not sure just yet - I need to think how I'd like to structure this content in general. Will let you know :)

docs/security-model.mdx
```diff
@@ -0,0 +1,55 @@
...
+   Ory's APIs available at `https://ory.example.org`.
+
+If you're interested in learning more about this topic, head over to
+[HTTP Cookies Explained](further-reading/http-cookies-explained.mdx)
+
+## Frequently Asked Questions
+
+Below we have collected some frequently asked questions about Ory's security model.
+
+### How do I solve CSRF Issues?
```
tomekpapiernik, line 38: What about this one? Still a WIP or do we need to write something for this paragraph?

docs/security-model.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+### What about Access Tokens / Refresh Tokens?
+
+Ory has an OAuth 2.0 / OpenID Connect product which can generate Access and Refresh Tokens. Unless
+[you need OAuth 2.0 / OpenID Connect](https://www.ory.sh/oauth2-openid-connect-do-you-need-use-cases-examples/), Ory's Security
+Model is a much easier solution to handle.
+
+### What about JSON Web Tokens?
+
+There is an [open issue to support JSON Web Tokens](https://github.com/ory/kratos/issues/2487) for Ory's Identity Service.
```
tomekpapiernik, line 55: I'd rather say that "currently, Ory doesn't support JWTs". When this is actually implemented we can easily update the docs.  

This way we avoid giving any hope / promising a feature (see the "no-promises policy" :) ).

src/sidebar.js
```diff
@@ -507,6 +509,12 @@ module.exports = {
       label: 'Reference',
       collapsed: false,
       items: ['reference/api']
+    },
+    {
+      type: 'category',
+      label: 'Further Reading',
+      collapsed: false,
+      items: ['further-reading/http-cookies-explained']
```
tomekpapiernik, line 517: I'd rather have this cookies information closer to the security model to make it more visible. Putting it into the "further reading" category is IMO equal to making it invisible :D

docs/security-model.mdx
aeneasr: Ok, we do want our users input on this though. So how about "let us know about your use case" and link to the issue?

src/sidebar.js
aeneasr: ok that makes sense to me

PR #824, opened 2022-05-24
docs: document --name flag for `ory create`

vinckr:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/docs/guides/cli/05_cli-basics.mdx
+++ b/docs/guides/cli/05_cli-basics.mdx
@@ -93,6 +93,8 @@ Run this command to create a new Ory Cloud project:
 ory create project
 ```
 
+Optional: Use the `--name` flag to specify the project name, for example `ory create project --name 'My Amazing Project'`.
+
 When the operation is successful, the system returns the details of the newly created project:
 
 ```text


tomekpapiernik:
TBH I'm not really huge on mentioning this flag :D I mean, when you simply run `ory create project`, in the next step the CLI asks you to provide the name of the project. I don't think that one method is quicker or more convenient than the other. 

But if you really feel that this should be in the docs, I'd suggest to make this comment a `:::tip` :)

You'd also have to rebase w/ master - there was a dead link which I fixed.

vinckr:
fair point, I just was looking for a way to do everything in one command and it took me a while to think of using `ory create project --help`.
just from the docs it dont think its SUPER clear you have these options (also for example supplying config with the [-c](url) flag). more advanced users will probably find this easily, and there is also the full reference. 
Maybe we should instead add a :::tip or a short paragraph that you can get a list of all options and flags with `ory create project --help`?

aeneasr:
So why not document `ory help create project`?

PR #827, opened 2022-05-24
add documentation for remote_json authenticator

vladr11:
Adds documentation for the `remote_json` authenticator.

ory/oathkeeper#841

### DIFF
--- a/docs/oathkeeper/pipeline/authn.md
+++ b/docs/oathkeeper/pipeline/authn.md
@@ -915,3 +915,85 @@ The request has been allowed! The subject is: "peter"
 
 In the background, this handler will fetch all JSON Web Key Sets provided by configuration key `authenticators.jwt.jwks_urls` and
 use those keys to verify the signature. If the signature can't be verified by any of those keys, the JWT is considered invalid.
+
+## `remote_json`
+
+The `remote_json` authenticator will forward the HTTP request (method, path, headers and body) to an upstream service. If the service returns `200 OK` and body `{ "subject": "...", "extra": {} }`, then the authenticator will set the subject appropriately.
+
+### `remote_json` Configuration
+
+- `service_url` (string, required) - The service to forward request method/path/headers/body to for authentication.
+- `preserve_path` (boolean, optional - defaults to `false`) - If set to `true`, any path in `service_url` will be preserved instead of replacing the path with the path of the request being checked.
+- `extra_from` (string, optional - defaults to `extra`) - A [GJSON Path](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) pointing to the `extra` field. This defaults to `extra`, but it could also be `@this` (for the root element), `session.foo.bar` for `{ "subject": "...", "session": { "foo": {"bar": "whatever"} } }`, and so on.
+- `subject_from` (string, optional - defaults to `subject`) - A [GJSON Path](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) pointing to the `subject` field. This defaults to `subject`. Example: `identity.id` for `{ "identity": { "id": "1234" } }`.
+- `method` (string, optional) - The method to pass to the authenticator service. If set, the method of the original request is overwritten with the specified method. If not set, the method of the original request is used.
+
+```yaml
+# Global configuration file oathkeeper.yml
+authenticators:
+  cookie_session:
+    # Set enabled to true if authenticator should be neabled and false to disable the authenticator. Defaults to false.
+    enabled: true
+
+    config:
+      service_url: https://auth-service-host
+```
+
+```yaml
+# Some Access Rule: access-rule-1.yml
+id: access-rule-1
+# match: ...
+# upstream: ...
+authenticators:
+  - handler: remote_json
+    config:
+      service_url: https://auth-service-host
+      preserve_path: false
+```
+
+```yaml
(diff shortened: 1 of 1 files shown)


docs/oathkeeper/pipeline/authn.md
```diff
@@ -915,3 +915,85 @@ The request has been allowed! The subject is: "peter"
...
+
+### `remote_json` Configuration
+
+- `service_url` (string, required) - The service to forward request method/path/headers/body to for authentication.
+- `preserve_path` (boolean, optional - defaults to `false`) - If set to `true`, any path in `service_url` will be preserved instead of replacing the path with the path of the request being checked.
+- `extra_from` (string, optional - defaults to `extra`) - A [GJSON Path](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) pointing to the `extra` field. This defaults to `extra`, but it could also be `@this` (for the root element), `session.foo.bar` for `{ "subject": "...", "session": { "foo": {"bar": "whatever"} } }`, and so on.
+- `subject_from` (string, optional - defaults to `subject`) - A [GJSON Path](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) pointing to the `subject` field. This defaults to `subject`. Example: `identity.id` for `{ "identity": { "id": "1234" } }`.
+- `method` (string, optional) - The method to pass to the authenticator service. If set, the method of the original request is overwritten with the specified method. If not set, the method of the original request is used.
```
piotrmsc, line 929: Open question : does it make sense to use the original request method as a fallback instead of the default method set to POST? For example, let's say we do a DELETE request and have set this authentication in our rule but haven't overridden the config method propoperty.

docs/oathkeeper/pipeline/authn.md
```diff
@@ -915,3 +915,85 @@ The request has been allowed! The subject is: "peter"
 
 In the background, this handler will fetch all JSON Web Key Sets provided by configuration key `authenticators.jwt.jwks_urls` and
 use those keys to verify the signature. If the signature can't be verified by any of those keys, the JWT is considered invalid.
+
+## `remote_json`
+
+The `remote_json` authenticator will forward the HTTP request (method, path, headers and body) to an upstream service. If the service returns `200 OK` and body `{ "subject": "...", "extra": {} }`, then the authenticator will set the subject appropriately.
```
piotrmsc, line 921: Could you extend a bit the description to point out that it aims to send both request body & headers? At least this is how I understood from the root issue description and would be nice to give here also a small note about the difference to `bearer_token` authenticator which you have pointed in the issue comment :)

piotrmsc (commented):
Thank you for providing the docs for changes you'd like to introduce in oathkeeper :)  I went through your code PR and have two open questions :)

docs/oathkeeper/pipeline/authn.md
vladr11: I think it makes sense because it would be the only way to pass the original method. Now it is true that usually an authentication server does not perform the authentication based on the request method, but I believe that it would be useful to cover such a case as well.

Another possibility would be to make the default behaviour to send a POST request to the authn server if no method is specified, and have another flag that enables the users to simply forward the original method as well, which by default is disabled, so one needs to explicitly enable the current behaviour which preserves the method.

piotrmsc: >Another possibility would be to make the default behaviour to send a POST request to the authn server if no method is specified, and have another config flag that enables the users to simply forward the original method as well, which by default is disabled, so one needs to explicitly enable the current behaviour which preserves the method.

Makes sense to me

docs/oathkeeper/pipeline/authn.md
vladr11: Added a note to point out the difference between the `remote_json` & `bearer_token` authenticators

docs/oathkeeper/pipeline/authn.md
```diff
@@ -915,3 +915,86 @@ The request has been allowed! The subject is: "peter"
...
+- `subject_from` (string, optional - defaults to `subject`) - A [GJSON Path](https://github.com/tidwall/gjson/blob/master/SYNTAX.md) pointing to the `subject` field. This defaults to `subject`. Example: `identity.id` for `{ "identity": { "id": "1234" } }`.
+- `method` (string, optional) - The method to pass to the authenticator service. If set, the method of the original request is overwritten with the specified method. If not set, the method used will be `POST` by default, or the original request method if the `use_original_method` fiels is set to `true`.
+- `use_original_method` (boolean, optional - defaults to `false`) - If set to `true`, the request to the authentication service will be performed using the original request method; otherwise, the request will be performed using the method `POST`. This field is ignored if the `method` field is set.
+
+```yaml
+# Global configuration file oathkeeper.yml
+authenticators:
+  cookie_session:
```
cjhines, line 935: Is `cookie_session` instead of `remote_json` here intentional?

vinckr:
added the upstream label to make clear this is waiting for the PR in oathkeeper.

aeneasr:
upstream is merged

docs/oathkeeper/pipeline/authn.md
aeneasr: Nice catch!

aeneasr:
```suggestion
  remote_json:
```

PR #833, opened 2022-05-27
docs: move API and SDK

aeneasr:
Move API and SDKs to top-level navbars which is very common in product docs because they are easy to find. See the Stripe example:

Screenshot 2022-05-27 at 17 20 34


aeneasr:
@tomekpapiernik resolved the conflicts, this can be merged now

PR #843, opened 2022-05-31
fix: grammar on security model

Benehiko:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).


docs/security-model.mdx
```diff
@@ -85,6 +85,12 @@ traffic (Kubernetes, OpenShift, ...).
 
 ## HTTP Cookies Explained
 
+:::tip
+
+Read our extensive blog post exploring [authentication and modern software](https://www.ory.sh/auth-and-modern-software/)
```
aeneasr, line 90: I think the blog article is great. I think we should keep this section strictly about "HTTP Cookies Explained". What about adding this article to a "further/recommended reading" section at the end?

docs/security-model.mdx
```diff
@@ -11,11 +11,11 @@ To achieve that, Ory manages:
 - tokens
 - cookies
 
-While this is not a common approach that might require a shift in thinking and some getting used to, it's implemented so that you
+Although it is an uncommon approach and might require a shift in thinking, Ory has implemented it so that you
```
tomekpapiernik, line 14:
```suggestion
Although it is an uncommon approach that might require a shift in thinking, it's implemented so that you
```

I'd like to limit the use of "Ory" in writing where possible. We're in the documentation for Ory after all, so no need to use the company that often :)

tomekpapiernik (changes requested):
Some good changes here, I'd love to see a few things tweaked though :)

docs/security-model.mdx
Benehiko: I have been reading over the article a bit more and thought it might make sense to have this at the top right after introducing the topic. The reason being is that this docs article does not really explain all the differences between cookies and tokens or what a session really is. But we do mention that we manage it in the introductory paragraph. I can have the `tip` there.

docs/security-model.mdx
```diff
@@ -3,69 +3,89 @@ id: security-model
...
+by managing `sessions`, `tokens` and `cookies`.
 
-To achieve that, Ory manages:
+Although it is an uncommon approach that might require a shift in thinking, it's
+implemented so that you don't have to worry about refreshing tokens or deciding
+whether to store them in `localStorage` or `document.cookies`.
 
-- sessions
-- tokens
-- cookies
```
tomekpapiernik, line 12: Oh, what happened here? The list was there to bring attention to the three elements Ory manages. 
Also, these weren't put into `code` bc. in this context it doesn't make sense IMO - these are not used as parameter names or code excerpts, but rather as proper names of elements of some system. Could you revert to the way it was before? 

Also, what's up with the line break? Lines look way shorter than the 130 char. limit we have defined.

tomekpapiernik (changes requested):

docs/security-model.mdx
tomekpapiernik: @Benehiko Well it's just an introductory sentence, we're explaining something, giving ppl an idea of what they can expect. Such general statements are not perfect, sure, but they're not the devil ;) So as long as they're not misleading I don't see a problem with using them sparingly.

Not to point fingers, but the `heavy lifting` sentence you wanted to use is just as verbose as this one haha :D

docs/security-model.mdx
```diff
@@ -3,30 +3,35 @@ id: security-model
 title: Security Model
 ---
 
-The security model used by Ory is designed to deliver the most secure and hassle-free solution to all users.
-
-To achieve that, Ory manages:
+Orys' security model delivers the most secure and hassle-free solution to users by managing:
```
tomekpapiernik, line 6:
```suggestion
The security model used by Ory delivers the most secure and hassle-free solution to users by managing:
```

`Orys'` is a possessive form of a plural noun, which is not correct. Let's stick to the previous approach, k?

tomekpapiernik (approved):

PR #847, opened 2022-06-02
Remove prettier

zepatrik:
See commits.


docs/cli/ory-tunnel.md
```diff
@@ -12,7 +12,7 @@ To improve this file please make your change against the appropriate "./cmd/*.go
 
 ## ory tunnel
 
-Tunnel Ory on a subdomain of your app or a seperate port your app's domain
+Tunnel Ory on a subdomain of your app or a separate port your app's domain
```
tomekpapiernik, line 15: Since the CLI docs are autogenerated, this will be changed to "separate" the next time CLI docs are generated until it's fixed in the code.

docs/guides/social-signin/microsoft.mdx
```diff
@@ -123,23 +124,23 @@ Follow these steps to add Microsoft as a social sign-in provider to your Ory Clo
...
 5. Add the social sign-in provider configuration to the downloaded config. Add the Jsonnet snippet with mappings as a Base64
    string or provide an URL to the file.
 
-  ```yaml
-  selfservice:
-    methods:
-      oidc:
-        config:
-          providers:
+```yaml
```
tomekpapiernik, line 138: The formatting tool removes the indents I make to enforce proper formatting for lists. Is there any way we could stop that and still have the line break guard active? See the screenshot for a visual comparison.

If there's no way to do that then please exclude the social sign in docs from this pr. I just spent some time fixing the formatting in these docs and this change basically screws most of that up lol

image

docs/cli/ory-tunnel.md
zepatrik: Ah I just used search&replace.

docs/guides/social-signin/microsoft.mdx
zepatrik: In yaml
```yaml
foo:
- bar
- baz
```
and
```yaml
foo:
  - bar
  - baz
```
is the same? Why would we have to use your indentation? In my opinion the variant enforced by prettier even makes more sense, because the array is under the key.

aeneasr: Thank you for paying close attention to this! :) 

In YAML you can write lists like the following and they are all equivalent.

```
foo:
  - bar
  - baz
```

```
# this one you see the most often
foo:
- bar
- baz
```

```
foo: ['bar', 'baz']
```

Hope this helps

zepatrik:
TL;DR
It is not possible to use prettier with the docusaurus mdx parser, it just breaks the rendering as seen in the screenshot. Therefore I removed prettier now :shrug:

tomekpapiernik (approved):
I'm good with prettier not affecting .md / .mdx 👍🏻

aeneasr:
I missed that the formatting is now completely gone, I thought we're only updating some formatting rules. I don't think the formatting tool chain should be removed unless there is a technical issue that prevents us from using it.

aeneasr:
> TL;DR It is not possible to use prettier with the docusaurus mdx parser, it just breaks the rendering as seen in the screenshot. The broken part is that the code blocks are not part of the list items anymore, not the content of the code block. Therefore I removed prettier now 🤷
> 
> image

Ah, I missed that as well.

package.json
```diff
@@ -76,6 +70,8 @@
     "@swc/core-darwin-x64": "^1.2.139"
   },
   "devDependencies": {
+    "ory-prettier-styles": "1.2.0",
+    "prettier": "2.6.2",
```
kevgo, line 74: These two dependencies are already production dependencies. Why are they added as devDependencies as well?

kevgo (commented):
@zepatrik @aeneasr @tomekpapiernik a better workaround than disabling all formatting is to [disable Prettier for only the problematic document parts](https://prettier.io/docs/en/ignore.html#range-ignore). Ideally these range-ignores contain a comment refering to the GitHub issue that tracks this Prettier bug. If such a bug report doesn't exist, please file one so that the Prettier folks can fix them.

tomekpapiernik:
If you can disable this behavior for `/docs` only then I'm all for that. 
I don't feel like going over all lists in this repository and changing indentation depth is something worth burning time on. 

> Gatsby works differently and considers content with less indentation than its surrounding list a part of that list. These ambiguities are one more reason to format our source code precisely and unambiguously with a tool instead of manually.

Why do you mention Gatsby? Am I missing something?

tomekpapiernik:
Also, doesn't the change in this PR affect `/docs` only?

PR #849, opened 2022-06-02
Fix config parameter name and curl method

kszafran:

The configuration parameter name was wrong in the documentation, and the `curl` example was missing `-X PATCH` (i.e. it would send a GET request).

I'd also note that the parameter name is somewhat unfortunate in my opinion. I was dead certain that if I set it to e.g. 1h, then I can refresh a session once an hour has passed since I logged it. But it actually means that the session can be refreshed 1 hour before the session expires... I see that internally the name `RefreshMinTimeLeft` is used, and it's way more clear.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/docs/guides/session-management/refresh-extend-session.mdx
+++ b/docs/guides/session-management/refresh-extend-session.mdx
@@ -41,7 +41,7 @@ import TabItem from '@theme/TabItem'
 <TabItem value="curl">
 
 ```shell script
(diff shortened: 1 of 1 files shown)


aeneasr (approved):
🙏 thank you so much

PR #864, opened 2022-06-13
feat: Configure Permission Service using CLI guide

zepatrik:
closes https://github.com/ory-corp/cloud/issues/2522

### DIFF
--- a/docs/guides/cli/basics.mdx
+++ b/docs/guides/cli/basics.mdx
@@ -1,6 +1,7 @@
 ---
 id: cli-basics
 title: Basic Functionality
+sidebar_position: 5
 ---
 
 To get a full list of commands available in Ory CLI, run this command:
--- a/docs/guides/cli/config-identity-service.mdx
+++ b/docs/guides/cli/config-identity-service.mdx
@@ -1,6 +1,7 @@
 ---
 id: identity-cli
 title: Configure Identity Service
+sidebar_position: 15
 ---
 
 Using the Ory CLI, you can quickly adjust the configuration of the Ory Cloud Identity Service and apply config from the Cloud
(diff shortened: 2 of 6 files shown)


docs/guides/cli/basics.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 id: cli-basics
 title: Basic Functionality
+sidebar_position: 5
```
zepatrik, line 4: Changing file content is better to review than changing file names IMO. I can also revert this, but we did not use the number convention elsewhere so this is also more consistent.

tomekpapiernik: Changing filenames broke two relative links which causes the website build to fail. Logs: 

```
Error: Docs markdown link couldn't be resolved: (../guides/cli/10_config-with-cli.mdx) in /docs/docs/concepts/personal-access-token.mdx for version current
    at Array.forEach (<anonymous>)
Error: Docs markdown link couldn't be resolved: (./cli/20_proxy-and-tunnel.mdx) in /docs/docs/guides/custom-domains.mdx for version current
    at Array.forEach (<anonymous>)
```

tomekpapiernik: The system where the document order is determined by filenames is something I'm trying out and I really like it so far. The benefits are: 
- the `sidebars` file is leaner
- you can see the correlation between the left navigation and the files in the directory at the first glance
- when you start messing with the file order you don't have to open any other files to know what to do - everything is visible at a glance without having to inspect the metadata section of the docs

I'm slowly implementing this system as I update/refactor/create new documents. For know, I think that this is the way to go. 

Please revert filenames to their initial state. This will fix your failing build.

docs/guides/cli/config-identity-service.mdx
```diff
@@ -54,7 +55,7 @@ ory get identity-config <project-id> --format yaml > identity-config.yaml
 
 ## Import Configuration to Ory Cloud
 
-To start using using the configuration from a self-hosted Ory Kratos instance in the Ory Cloud Identity Service, import the
+To start using the configuration from a self-hosted Ory Kratos instance in the Ory Cloud Identity Service, import the
```
tomekpapiernik, line 58: 🙆🏻

tomekpapiernik (changes requested):
The document you added looks okay - it's basically the same as the Identity Service. 
Please revert back to the old file naming structure to fix the builds and make your TW happy.

PR #902, opened 2022-07-14
docs: create getting started section

tomekpapiernik:
## Checklist

## Further comments

As a part of the planned restruct of the docs, this PR creates a new "getting started" section. This section is made up mostly of the existing "protect page with login" docs arranged into categories. A new document for React Native has been added.


docs/getting-started/integrate-auth/react-native.mdx
```diff
@@ -0,0 +1,161 @@
+---
+id: react-native
+title: Integrate Authentication into React Native
+sidebar_label: React Native
+---
+
+import CodeFromRemote from "@theme/CodeFromRemote"
```
aeneasr, line 7: Please wrap all MDX code in an [mdx code block](https://github.com/ory/docs/#import-markdown) in this document!

docs/getting-started/integrate-auth/react-native.mdx
```diff
@@ -0,0 +1,161 @@
+---
```
aeneasr, line 1: Have you tried running this guide on your machine?

aeneasr (commented):
Thank you! This restructuring looks much better already! I just have two or three minor things.

On a side note, it's usually best practice to create rewrites for *all* resources you move, including videos and pictures. That's important for both SEO (when doing image search) as well as external links to these resources.

If you move many files from directory A to B, you only need one rule, see: https://vercel.com/support/articles/can-i-redirect-from-a-subdomain-to-a-subpath

docs/getting-started/integrate-auth/react-native.mdx
tomekpapiernik: I just checked it out and unfortunately, something seems to be wrong. I mean I can get the app to "boot" in the web mode, but other than clicking the image button there isn't anything else I can do. No fields to provide user email/password are shown. I also found that:

- By default, the app uses a playground project address that I guess is very old. The address is: `https://demo.tenants.staging.oryapis.dev/api/kratos/public`. I think we should change this to our current playground project URL. (I replaced it with my project's SDK URL and that didn't help). 
- Running `npm run web` works, but running the app in the web env. isn't possible from the Expo dash since `"web" is not included in the project app.config.js or app.json "platforms" array`. Easy fix. 

I wasn't able to run the app on simulated Android (duh) or iOS (Xcode is needed and doesn't want to install for some reason). 

After that fiasco I went back to the original blog post and ran every installation command. Still doesn't work. 

@aeneasr do you think you can have a look at the app and debug? If not, I can ask someone else from the team on Monday :)

aeneasr: Ok thank you for checking, I probably won't be able to look into this this week. Someone with a bit of expertiese in this should look into it though, it's a bit tricky to set up XCode etc.

Should we remove this doc from the guide and move it to another PR so that we can merge the restructure and add the native guide once it works?

tomekpapiernik: Yes, let's do this. I'll move this doc to a separate PR 👍🏻

tomekpapiernik:
> If you move many files from directory A to B, you only need one rule, see: https://vercel.com/support/articles/can-i-redirect-from-a-subdomain-to-a-subpath

Thanks for the tip :) It seems that although the relative path changes, the absolute path on the static site remains the same. I came to that conclusion by inspecting the video player for the next.js document on our current master (where it lives under `/guides`) and on the vercel-deployed preview (where it lives under `getting-started`). The URI in both cases is `/docs/assets/medias/screencast-2e87c91cbac48fd00c7721bab40ac4fd.webm`.

aeneasr:
👍

PR #918, opened 2022-07-20
docs: kratos deploy guide fixes

vinckr:
Fixes the Kratos deploy guide with the same improvements as suggested for the hydra guide.

### DIFF
--- a/docs/reference/api.json
+++ b/docs/reference/api.json
@@ -274,7 +274,8 @@
         "type": "object"
       },
       "SubscriptionStatus": {
-        "title": "SubscriptionStatus is the list of allowed values for the subscription's status.",
+        "description": "For `collection_method=charge_automatically` a subscription moves into `incomplete` if the initial payment attempt fails. A subscription in this state can only have metadata and default_source updated. Once the first invoice is paid, the subscription moves into an `active` state. If the first invoice is not paid within 23 hours, the subscription transitions to `incomplete_expired`. This is a terminal state, the open invoice will be voided and no further invoices will be generated.\n\nA subscription that is currently in a trial period is `trialing` and moves to `active` when the trial period is over.\n\nIf subscription `collection_method=charge_automatically` it becomes `past_due` when payment to renew it fails and `canceled` or `unpaid` (depending on your subscriptions settings) when Stripe has exhausted all payment retry attempts.\n\nIf subscription `collection_method=send_invoice` it becomes `past_due` when its invoice is not paid by the due date, and `canceled` or `unpaid` if it is still not paid by an additional deadline after that. Note that when a subscription has a status of `unpaid`, no subsequent invoices will be attempted (invoices will be created, but then immediately automatically closed). After receiving updated payment information from a customer, you may choose to reopen and pay their closed invoices.",
+        "title": "Possible values are `incomplete`, `incomplete_expired`, `trialing`, `active`, `past_due`, `canceled`, or `unpaid`.",
         "type": "string"
       },
       "Time": {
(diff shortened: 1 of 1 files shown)


docs/kratos/guides/deploy-kratos-example.mdx
```diff
@@ -127,24 +130,24 @@ Type "help" for help.
...
 ```bash
 useradd -s /bin/false -m -d /opt/kratos kratos
 ```
 
-Create folders to hold the installation data and configuration files.
+2. Create folders to hold the installation data and configuration files.
 
 ```bash
 mkdir /opt/kratos/{bin,config}
 ```
```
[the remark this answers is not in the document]
vinckr: ah great point. I indented them, only the "install kratos/hydra <codeblock>" did not want to be indented, or rather prettier would format it in a way that broke everything. 

alas we cant indent dynamic codeblocks in lists atm, but I think that is OK.

docs/kratos/guides/deploy-kratos-example.mdx
```diff
@@ -63,405 +59,414 @@ apt-get install nodejs jq unzip -y
...
-  }
-}
-```
+   ```bash
+   19191 ?        Ssl    0:00 /opt/kratos/bin/kratos -c /opt/kratos/config/kratos.yml serve
+   19206 ?        Ss     0:00 postgres: 12/main: kratos kratos 127.0.0.1(36094) idle
+   ```
+
+7. Check if it's accessible (if your DNS record is not set yet (it takes up to 24h to propagate), you can also send the request
+   here directly to your Droplet IP)
```
tomekpapiernik, line 274: Lol parenthesis in a parenthesis whaaaaat

docs/kratos/guides/deploy-kratos-example.mdx
```diff
@@ -6,55 +6,51 @@ title: Deploy to production
...
+First, upgrade all packages in your Droplet.
 
 ```bash
 apt-get update && apt-get upgrade -y
 ```
 
-The default version of [Node.js](https://nodejs.org/en/) is v10.19.0 on Ubuntu 20.04, we need to install a newer version. You can
-use the following script to install Node.js 16.x on a Ubuntu system.
+Since the default version of [Node.js](https://nodejs.org/en/) is outdated we need to install a newer version. You can use the
+following script to install Node.js 16 on a Ubuntu system.
```
tomekpapiernik, line 53: This sounds a bit weird. Maybe re-phrase it to say that the user needs a specific version of Node.js to do all the things we talk about in this guide? I mean, 16.14.2 is bound to be outdated very soon (as a matter of fact 16.16.0 seems to be the newest v.) Also, would be great if you could get rid of the "we" form ;)

docs/kratos/guides/deploy-kratos-example.mdx
```diff
@@ -63,405 +59,414 @@ apt-get install nodejs jq unzip -y
...
+```bash
 INFO[2022-05-10T20:51:56Z] TLS has not been configured for public, skipping  audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.3
 INFO[2022-05-10T20:51:56Z] Starting the public httpd on: 0.0.0.0:4433    audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.3
 INFO[2022-05-10T20:51:56Z] TLS has not been configured for admin, skipping  audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.3
-INFO[2022-05-10T20:51:56Z] Starting the admin httpd on: 0.0.0.0:4434     audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.3```
-````
+INFO[2022-05-10T20:51:56Z] Starting the admin httpd on: 0.0.0.0:4434     audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.3
+```
 
 ## Run Ory Kratos using Systemd
```
tomekpapiernik, line 219: Please make the capitalization of `Systemd` uniform across the doc. Here it's capitalized, in other places it's not. 
Shouldn't start with a capital letter IMO: https://systemd.io/

docs/kratos/guides/deploy-kratos-example.mdx
```diff
@@ -470,245 +475,238 @@ nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
...
 
-Save changes and restart Ory Kratos:
+3. Save changes and restart Ory Kratos:
 
-```
-systemctl restart kratos
-```
+   ```
+   systemctl restart kratos
+   ```
```
tomekpapiernik, line 709: Something weird going on here, pls have a look at the preview and do a fixaroo

image

tomekpapiernik (changes requested):

docs/kratos/guides/deploy-kratos-example.mdx
tomekpapiernik: Same comment applies to the Hydra guide.

vinckr: You dont need a specific version to do the guide - you just need the 16.x version installed with the script below, it will install the latest version. 

We had the same discussion on the update to hydra guide 😉

docs/kratos/guides/deploy-kratos-example.mdx
vinckr: yes you are right, it should be lowercase everywhere, thanks

docs/kratos/guides/deploy-kratos-example.mdx
tomekpapiernik: Oh, ok, totally forgot about it

tomekpapiernik (approved):

PR #945, opened 2022-08-09
docs: add a guide for migrating from Hydra v1 to Hydra v2

grantzvolsky:
## Checklist


tomekpapiernik:
Hey @grantzvolsky, I checked out this document and I have some doubts whether this should be a separate document at all. 
From what I can see, there's barely anything "new" here when I compare the process to the general upgrade guide for Hydra. 

I feel that any specifics of the process could be added to the existing "upgrade" document - there are just one or two things that look like they're Hydra v2-specific. 
Is deploying to a "dry-run env" something that's absolutely necessary? I get why you included this in the document, but truth be told - this is something the users can do if they want to, but if they're feeling adventurous they can apply the upgrade on "prod", right? Hence, this looks more like a suggestion to me. 

This document is also very inconsistent when it comes to the terminology it uses. On one occasion it's "upgrade", then it's "migration". From my point of view, bringing up software to a new version is an upgrade, while moving data between containers is a migration. So in this case, you upgrade Hydra to v2 and a part of this process is migrating the SQL DB. 

Also, you say that there's no downgrade path from v2. Is this by any chance true for other versions?

grantzvolsky:
@tomekpapiernik 

> From what I can see, there's barely anything "new" here when I compare the process to the general upgrade guide for Hydra.

It is positive that we have been able to make the migration to Hydra 2.x automatic, as it was initially unclear whether this would be possible. We achieved this with some rather heavy migrations, so while it is possible to carry out the migration without manual intervention, it does require a great deal of care.

> I feel that any specifics of the process could be added to the existing "upgrade" document - there are just one or two things that look like they're Hydra v2-specific.

Eventually, yes, that is a good idea to update the existing upgrade document with bits of the process presented here.

> This document is also very inconsistent when it comes to the terminology it uses. On one occasion it's "upgrade", then it's "migration".

Upgrade refers to code. Migration refers to data. Please highlight the usages where this distinction isn't clear.

> Is deploying to a "dry-run env" something that's absolutely necessary?

Different deployments will have different requirements and preferences. The point of this guide is to cover the essentials that minimize the risk of something going wrong. However, due to the diversity of environments in which Hydra is deployed, the guide cannot be prescriptive. Furthermore, the reason for this guide to be in a separate document is to make it easier to collect and integrate feedback specifically for the 2.x release. As stated above, the migrations included here are quite heavy.

> Also, you say that there's no downgrade path from v2. Is this by any chance true for other versions?

One should not generally rely on down migrations, but v2 is an exception in that it doesn't offer them.

grantzvolsky:
@aeneasr I expect that we'll add more information after we get feedback from beta testers, but for now this is it. Any feedback is welcome.

By the way, should I close this PR and add the changes to https://github.com/ory/docs/pull/874 instead?

docs/hydra/guides/migrate-v2.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+:::warning
+
+This is an offline migration. You will need to take Hydra offline for the duration of the migration.
+
+:::
+
+:::warning
+
+There is no down migration from v2 to v1.
```
tomekpapiernik, line 26: The term "down migration" is not recognized by dictionaries and gets about 154,000 google results. Please use "downgrade" which is a widely used and understood term that gets about 28,400,000 google results. 

[Wikipedia for "downgrade"](https://en.wikipedia.org/wiki/Downgrade)

docs/hydra/guides/migrate-v2.mdx
```diff
@@ -0,0 +1,110 @@
...
+3. Take your deployment offline.
+
+4. Backup your data.  
+   Be prepared to go back online with the original database and Ory Hydra v1.
+
+5. Switch your deployment to the version successfully tested in the dry run.
+
+6. Run [`hydra migrate sql`](../cli/hydra-migrate-sql.md) to apply the SQL migrations.
+
+7. Run the [sanity checks](#known-issues-limitations-and-sanity-checks).
```
tomekpapiernik, line 80: There is a single "sanity check". The section you're referring to contains a single simple query and it could be made an integral part of the instruction.

docs/hydra/guides/migrate-v2.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+1. Check out [Known issues, limitations, and sanity checks](#known-issues-limitations-and-sanity-checks).
+
+2. Backup your data.
+
+3. Deploy a dry-run environment with the backed-up data.
+
+4. Take the dry-run environment offline.
+
+5. Apply the Hydra v2 patch to the dry-run environment.
```
tomekpapiernik, line 54: Isn't applying the patch the essence of upgrading? Does it move the data between DBs in any way?

docs/hydra/guides/migrate-v2.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+6. Run [`hydra migrate sql`](../cli/hydra-migrate-sql.md) to run the SQL migrations.
+
+7. Start the dry-run environment.
+
+8. Test the dry-run environment.  
+   Make sure to test your deployment before you go live. Hydra releases are tested, but each deployment is different and it is
+   critical to have your own tests.
```
tomekpapiernik, line 62: This is something more suitable for an admonition rather than the "regular content" of the guide IMO. You'd hope that a bunch of cool dawgs like us would test the new version of the software before it's released :D

docs/hydra/guides/migrate-v2.mdx
```diff
@@ -0,0 +1,110 @@
+---
+id: migrate-v2
+title: Migrate from Hydra v1 to v2
```
tomekpapiernik, line 3: This is not a migration but an upgrade that requires data migration.

tomekpapiernik (changes requested):
@grantzvolsky I highlighted some of the issues I see with the document. In general, I think it doesn't need to be a separate document as the general upgrade process isn't different from the generally established path that's applicable for other versions. 

As such, I'd recommend adding a dedicated section that focuses on the specifics of the v1 to v2 upgrade to the existing document about upgrades. 

Another recommendation/request is to make the document more unambiguous and to the point so that it's easier to follow. The documentation has to make clarity and brevity a priority - people with different levels of language proficiency are going to be looking at this content so it's our duty to make sure that we're not making things harder for them in any way. 

I hope this helps :) Should you have more questions, please reach out to me directly. I'll be happy to help!

docs/hydra/guides/migrate-v2.mdx
grantzvolsky: I'm afraid I disagree with this suggestion. "downgrade" is not the same as "down migration", and the intended audience will understand it. The meaning of "down migration" can be understood from the common term itself.

docs/hydra/guides/migrate-v2.mdx
grantzvolsky: This section is designed to encourage people to contribute their own checks. The language is deliberately not meant to sound authoritative in order to encourage people to contribute.

docs/hydra/guides/migrate-v2.mdx
grantzvolsky: The patch I'm referring to is the code that users write to upgrade their deployment. Whether or not it moves any data depends how one implements it. e.g. it would be the commit that upgrades to Hydra 2.0 in the Ory Cloud repository.

docs/hydra/guides/migrate-v2.mdx
grantzvolsky: True, good catch :)

grantzvolsky:
@tomekpapiernik I've updated the title & added some comments.

docs/hydra/guides/migrate-v2.mdx
tomekpapiernik: 👍🏻 okay, I think I understand things better now. This term can stay in the doc.

tomekpapiernik:
@grantzvolsky I talked to @piotrmsc yesterday and I think I understand the process now. 
From what I understood, it all boils down to:

1. Download the new Hydra CLI version. 
2. Run the `[hydra migrate sql](https://www.ory.sh/docs/hydra/cli/hydra-migrate-sql)` command to migrate the DB to the "new, v2 format". 
3. Run the newest version of Hydra, however you (the user) do it in your setup. 

This process is described in the existing document here: https://www.ory.sh/docs/hydra/guides/upgrade

If you feel the v1 to v2 migration needs any additional procedures other than these, please add them to the existing documents. IMO adding one more document that's basically the same content but with just a little-to-no twist won't do users no good and will rather unnecessarily confuse them.

PR #954, opened 2022-08-17
docs: move api and sdk reference to top navbar

tomekpapiernik:
This PR elevates API reference and SDKs to the top navigation bar. 
Since the overall API reference lives in the top navigation now, I removed it from the sidebar to keep things nice and tidy. 

## Checklist


vinckr (approved):
Not 100% sure about the API reference (since its just cloud now) but this is still better than now :) 

thanks! we can probably also much improve the SDK content, but like this its not to be missed.

we should rename the sdk categories with Identity Service (Kratos) etc. 
but I can take care of that tomorrow

🚢

PR #955, opened 2022-08-17
fix: sdk overview sidebar

vinckr:
add an overview sdk page for all four services and put it in sidebar on the top

## Checklist

- [x] I have added necessary documentation within the code base (if appropriate).


kevgo (approved):
Nice improvement! 

The Oathkeeper SDK sidebar contains only a single entry "overview". Are you planning on adding more pages to it or do we want to get rid of it?

image

vinckr:
> The Oathkeeper SDK sidebar contains only a single entry "overview". Are you planning on adding more pages to it or do we want to get rid of it?

No immediate plans, but yea longer term we should have some SDK examples for Oathkeeper as well. Maybe we can just ask the users in Slack or on here if they can provide some example code they are using in their own projects, WDYT? 

But yea we can start with the Oathkeeper Go SDK example soon !

kevgo:
If we have no immediate plans, can we get rid of the single "Overview" sub-entry for now and make the "Ory Oathkeeper" menu entry show the page for the Oathkeeper SDK directly?

kevgo (approved):
You are a Da-Vinci of Docusaurus sidebars!

kevgo:
Gaa, can't approve...

tomekpapiernik (changes requested):
Not a fan of the changes here :D 
1st of all, let's not use the `sidebar_position:` property for ordering files. Let's stick to either ordering through filenames or explicit ordering in sidebars.js for directories that don't use the "autogenerated" option. 

I was expecting to see an overview document that's separate from all other categories but that isn't the case. From what I can see that's because you added the `hydra/sdk/index` to a directory that's `autogenerated` so instead of this file showing up as an item separate from all categories it's simply another document in Hydra. To combat that, place the file in a different location and reference it directly, before the autogenerated sections.

tomekpapiernik:
```
sdk: [
   "SDK-OVERVIEW-FILE-HERE"
    {
      "Ory Kratos": [
        {
          type: "autogenerated",
          dirName: "kratos/sdk",
        },
      ],
    },
    {
      "Ory Hydra": [
        {
          type: "autogenerated",
          dirName: "hydra/sdk",
        },
      ],
    },
    {
      "Ory Keto": [
        {
          type: "autogenerated",
          dirName: "keto/sdk",
        },
      ],
    },
    {
      "oathkeeper/sdk",
    },
  ],
}
```

Something like this should work. :) You can put the SDK overview right in /docs alongside the security model document, for example.

vinckr:
Fair point with the ordering will change that. 

As for 
>I was expecting to see an overview document that's separate from all other categories



Right now - status quo - there is an overview document with all sdks for keto, oathkeeper and kratos linked. but not for Hydra. 
The point of this PR is to add the hydra overview doc to the SDK page. 
Screenshot 2022-08-18 at 10 04 34

A general overview page for all SDKs would also be possible, but will take a bit longer probably.

As for the Hydra file, I just changed the naming/location etc. so its consistent with the other files. So all sdk overview files are the same now. We can change all of them, but we should not have different files there for different services, right?

Tbh I dont get [this comment](https://github.com/ory/docs/pull/955#issuecomment-1219138990), do you want to get rid of all the overview files and merge it into on big? 
Not sure why though, people ususally look for the SDK for the service they are using atm? 
I know we have plans to merge the SDKs longer term, but I think we are still far off that. 

I will push the ordering change in a minute, but maybe you can just push on this branch with what you have in mind @tomekpapiernik or elaborate a bit so I understand what you are looking for?

tomekpapiernik:
Okay, now that I read this comment: 
> The point of this PR is to add the hydra overview doc to the SDK page.

Things make more sense :) 

With the original description that said: 
> add an overview sdk page for all four services and put it in sidebar on the top

it sounded like you were meaning to add a "master overview" sort of thing.

vinckr:
> Regarding filenames vs ordering, please use `01_`, `05_`, `10_`, etc. This will make it easier to stick a file between positions 1 & 2. In fact, you'll be able to add 3 files before you have to rename anything. Make sure to apply the names to all files in the directory. Remove the `sidebar_position` front matter then, you won't need it.

Thanks, should be fine now. Not too many files in there at the moment anyway 😢

tomekpapiernik:
This is the structure I'm looking for:

image

This way you can easily add 3 files in between other files w/o the need to rename them. For example: 

image

docs/kratos/sdk/01_index.md
```diff
@@ -2,6 +2,7 @@
 id: index
 title: SDKs
 sidebar_label: Overview
+sidebar_position: 1
```
tomekpapiernik, line 5:
```suggestion
```

docs/keto/sdk/01_index.md
```diff
@@ -1,6 +1,7 @@
 ---
 id: index
 title: Overview
+sidebar_position: 1
```
tomekpapiernik, line 4:
```suggestion
```

tomekpapiernik (changes requested):

tomekpapiernik (approved):
Looks good now :)

PR #967, opened 2022-08-25
docs: explain insecure Keto CLI flag

hperl:
This PR quickly explains the usage of the new --insecure flags in the Keto CLI.

Blocked by https://github.com/ory/keto/pull/991

### DIFF
--- a/docs/keto/guides/expand-api-display-who-has-access.mdx
+++ b/docs/keto/guides/expand-api-display-who-has-access.mdx
@@ -59,7 +59,14 @@ directories:/photos#parent@(files:/photos/mountains.jpg#_)
 ```
 
 The user `maureen` now wants to manage `access` for the file `/photos/beach.jpg`. Therefore, the application uses the expand-API
-to get a tree of everyone who has access to that file:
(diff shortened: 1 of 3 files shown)


zepatrik (dismissed):
Lets merge this once the upstream Keto change is done.

tomekpapiernik (changes requested):
Don't be afraid to speak from the position of authority in the docs :) Replace "is necessary" with "you must" to stress the importance of this setting. 

"newer versions of the Keto CLI" sounds ambiguous - which version of the CLI you consider "newer"? Newer than what? Use specific version IDs for clarity.

Finally, instead of adding this information into parentheses, please add it as an admonition. If you feel that this information is very important, use `:::warning`. If not, `:::info` or `::note` should suffice.

tomekpapiernik:
@hperl Can you please finish this up? My comment still stands. Thanks.

hperl:
> @hperl Can you please finish this up? My comment still stands. Thanks.

Should be done now :tada:

tomekpapiernik (approved):

PR #979, opened 2022-09-01
docs: add before hooks

aeneasr:
For https://github.com/ory/kratos/pull/2706

### DIFF
--- a/docs/kratos/hooks/01_configure-hooks.mdx
+++ b/docs/kratos/hooks/01_configure-hooks.mdx
@@ -20,8 +20,11 @@ You can use hooks at specific points of select self-service flows:
 - `after` login: The hook runs when the user is successfully authenticated, before the system issues an Ory Session.
 - `before` registration: The hook runs when a registration flow starts.
 - `after` registration: The hook runs when a user successfully registers in the system.
+- `before` recovery: The hook runs when the user starts the recovery flow.
 - `after` recovery: The hook runs when the user successfully recovers their password.
+- `before` settings: The hook runs when the user starts the account settings flow.
 - `after` settings: The hook runs when the user successfully changes their account settings.
+- `before` verification: The hook runs when the user starts the verification flow.
 - `after` verification: The hook runs when the user verifies their account.
 
 ### Hooks for specific methods
--- a/docs/kratos/hooks/01_configure-hooks.mdx
+++ b/docs/kratos/hooks/01_configure-hooks.mdx
@@ -373,10 +373,10 @@ To do that, set `response.ignore` to `true` in the webhook config:
 ```yaml
 - hook: web_hook
   config:
-  // highlight-start
+    # highlight-start
     response:
       ignore: true
-  // highlight-end
+    # highlight-end
 ```
 
 #### Flow-interrupting webhooks


docs/kratos/hooks/01_configure-hooks.mdx
```diff
@@ -20,8 +20,11 @@ You can use hooks at specific points of select self-service flows:
 - `after` login: The hook runs when the user is successfully authenticated, before the system issues an Ory Session.
 - `before` registration: The hook runs when a registration flow starts.
 - `after` registration: The hook runs when a user successfully registers in the system.
+- `before` recovery: The hook runs when the user starts the recovery flow.
```
kevgo, line 23: Taking a step back from this particular tree and looking at the forest:

Is the name of this hook `before` or `before recovery`? If it's the former, this paragraph describes 4 other hooks with the same name. That's confusing. Also, the repetitive phrase ("The hook runs when ...") in each sentence makes this read somewhat clumsy and robotic. What about this shorter and more natural version?

```suggestion
- `before recovery` runs when the user starts the recovery flow.
```

Ideally this also contains a link to the page that explains how to start a recovery flow.

kevgo (approved):
The changes themselves are +1, one broader issue stood out while reviewing this. I'm happy to do this in a separate PR.

docs/kratos/hooks/01_configure-hooks.mdx
```diff
@@ -373,10 +376,10 @@ To do that, set `response.ignore` to `true` in the webhook config:
 ```yaml
 - hook: web_hook
   config:
-  // highlight-start
+    # highlight-start
     response:
       ignore: true
-  // highlight-end
+    # highlight-end
```
tomekpapiernik, line 382: I compared this with the current master and there seems to be no difference in how this is rendered. I used `//` for the line highlighting trigger based on the Docusaurus docs: 
https://docusaurus.io/docs/markdown-features/code-blocks#line-highlighting

So what you did here is just using the yaml-appropriate commenting syntax, correct? The magic comment `higlight-start/end` doesn't seem to care about it.

docs/kratos/hooks/01_configure-hooks.mdx
tomekpapiernik: This list shows users when they can trigger hooks:
image

`before` isn't a hook name, it defines the point at which the hook is triggered.

tomekpapiernik (approved):

docs/kratos/hooks/01_configure-hooks.mdx
aeneasr: 🤔 yeah the list is not very clear. It's not clear how you configure the individual hooks. Ideally, it would probably be linking to somewhere more extensive like the examples below. For this PR, I just wanted to get the new features in and follow the structure already set, but we're always glad and happy to get better documentation so any contribution and improvements are greatly appreciated :))

docs/kratos/hooks/01_configure-hooks.mdx
aeneasr: `//` is not a valid YAML comment, so code formatting is broken. YAML comments use hashtags `#`: https://www.tutorialspoint.com/yaml/yaml_comments.htm

docs/kratos/hooks/01_configure-hooks.mdx
tomekpapiernik: If you look at the list on its own then it doesn't tell you how to configure individual hooks of course. Please look at the complete document and you'll see that this section is meant to list the points at which hooks can be triggered. Examples are provided later on. Everything is easily accessible and scannable using the ToC (right-hand nav. in Docusaurus).

aeneasr: Ah, I probably missed that!

PR #1012, opened 2022-09-28
docs: corrects kratos example diff

gabeio:
127.0.0.1 should be replaced by accounts.example.com within the example diff.

N/a

## Checklist

## Further comments

Just a small change to correct the diff to match the rest of the surrounding diff within the example.

### DIFF
--- a/docs/kratos/guides/deploy-kratos-example.mdx
+++ b/docs/kratos/guides/deploy-kratos-example.mdx
@@ -571,8 +571,8 @@ You can use `ps ax | grep node` to confirm it is running.
    -  default_browser_return_url: http://127.0.0.1:4455/
    +  default_browser_return_url: https://accounts.example.com/auth/
      allowed_return_urls:
-   -    - https://accounts.example.com
-   +    - http://127.0.0.1:4455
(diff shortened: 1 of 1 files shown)


vinckr (commented):
Apologies maybe I did not get something, 
but would this have any effect? 
in both cases both return urls are allowed, right?

gabeio:
> in both cases both return urls are allowed, right?

This seems to be a diff explaining what should be removed and what should be added.

Screen Shot 2022-09-29 at 11 12 29

Above is the current look, notice that everything is shown as red is `127.0.0.1` and everything in green is `accounts.example.com`, I believe the idea is to show people to remove 127.0.0.1 from the example config provided and add their domain instead. So 127.0.0.1 should not be in the allowed return urls.

Sorry for not explaining that better in the original pr.

vinckr (approved):
I see! 
My bad I did not look too closely at the PR 😅 

Thanks for catching this! 

Was the guide helpful to you?

gabeio:
> Was the guide helpful to you?

Still working through it, but so far yes!

PR #1027, opened 2022-10-07
fix: oidc example code cloud

vinckr:
Adds an example to the oidc code boxes, since the method is enabled (for quite some time now ;))

also fixes package-lock so `npm ci` works again

### DIFF
--- a/docs/kratos/self-service/flows/user-login.mdx
+++ b/docs/kratos/self-service/flows/user-login.mdx
@@ -67,12 +67,22 @@ Two login methods are supported:
   ]}>
   <TabItem value="oc">
 
-Only the password method is enabled.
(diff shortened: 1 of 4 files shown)


docs/kratos/self-service/flows/user-login.mdx
```diff
@@ -67,12 +67,22 @@ Two login methods are supported:
...
+    oidc:
+      enabled: true
+  # ...
+```
 
   </TabItem>
   <TabItem value="oss">
 
-You can configure which methods to use in the Ory Kratos config:
+You can choose which methods to use in the Ory Kratos configuration:
```
kevgo, line 85: This reads much better!

kevgo (approved):

tomekpapiernik (approved):

PR #1029, opened 2022-10-11
docs: how to use hydra with cloudhsm

vinckr:
some tips and tricks how to use Hydra with AWS CloudHSM as shared by a community member. 
Please not that this was found out with painstaking reverse engineering and is __not__ documented properly in AWS docs.


docs/hydra/faq.md
```diff
@@ -10,6 +10,43 @@ removes or updates it!**
 
 <!-- toc -->
 
+## How to use Ory Hydra with AWS CloudHSM
```
aeneasr, line 13: Maybe move this to HSM docs directly?

aeneasr: I mean adding it here: https://www.ory.sh/docs/hydra/guides/hsm-support

vinckr: Oh my bad, I totally missed that document 🙈 
Will add it there thanks!

vinckr:
There is more to unpack in this document, but I removed some redundant tags and added a section for the AWS tips.

tomekpapiernik (approved):

PR #1030, opened 2022-10-11
docs: edit and rename technical advisory

tomekpapiernik:
This PR moves the recently added Technical Advisory document to a new Troubleshooting section in the Identity Management topic. This is a more fitting solution since the document describes steps required to fix an error in social sign-in behavior.

## Checklist


aeneasr:
I think it would be better to have a dedicated "troubleshooting" section for all of Ory. In the future, we will not differentiate any more as much as we do today between individual services, and the troubleshooting should probably be a global directory and menu item :)

tomekpapiernik:
> I think it would be better to have a dedicated "troubleshooting" section for all of Ory. In the future, we will not differentiate any more as much as we do today between individual services, and the troubleshooting should probably be a global directory and menu item :)

Are you sure? As you know, I'm very pro keeping things topical, so I advise having area-specific troubleshooting docs. I suspect that the number of documents of this kind will grow at a very slow pace so we'll have a very small number of documents relegated to a top-level section akin to identity management or getting started.  In the plans we've drawn up so far a section like this was never a top-level item either. Do you think we can merge this PR and maybe re-evaluate the approach in the future?

aeneasr:
Please keep this all in one section. We can always split if it becomes too much, right now it's just one or two documents. Thanks!

tomekpapiernik:
@aeneasr ok, done. Hope it's good to merge now ;)

PR #1036, opened 2022-10-13
fix: update keto prod guide

vinckr:
Updates the current "prod" guide for Keto to a "prepare for production" guide as we have in Hydra, with a similar structure - this should stop the bleeding until we create a proper "selfhosted" section for Keto. 

Also a few stylistic changes in the Hydra guide.

### DIFF
--- a/docs/keto/guides/production.md
+++ b/docs/keto/guides/production.md
@@ -11,13 +11,14 @@ Feel free to open a pull request when you have an idea how to improve this docum
 Ory Keto requires a production-grade database such as PostgreSQL, MySQL, CockroachDB. Don't use SQLite in production! Read more
 about [deployment fundamentals and requirements for Ory](https://www.ory.sh/docs/ecosystem/deployment).
 
-## Ory Keto Write API behind an API gateway
+## Ory Keto API behind an API gateway
 
 Although Ory Keto implements all Go best practices around running public-facing production HTTP servers, we discourage running Ory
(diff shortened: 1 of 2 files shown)


docs/keto/guides/production.md
```diff
@@ -1,23 +1,24 @@
...
-### Write API
+## Ory Keto Write API behind an API gateway
 
-Never expose the Ory Keto Write API to the internet unsecured. Always require authorization. A good practice is to not expose the
-Write API at all to the public internet and use a Zero Trust Networking Architecture within your intranet.
+Although Ory Keto implements all Go best practices around running public-facing production HTTP servers, we discourage running Ory
+Keto facing the public net directly. We strongly recommend running Ory Keto behind an API gateway or a load balancer. It's common
+to terminate TLS on the edge (gateway / load balancer) and use certificates provided by your infrastructure provider such as AWS
+CA for last mile security. A good practice is to not expose the Write API at all to the public internet and use a Zero Trust
+networking architecture within your intranet.
```
zepatrik, line 20: Actually all Keto APIs should be protected at any point!
Depending on the use-case even the Read API might expose too much information, e.g. I could check if you have permission to do sth, or could check who has permission to do sth, ...
Can you rephrase this section to reflect that?

zepatrik (changes requested):

zepatrik (approved):

PR #1054, opened 2022-11-02
docs: update sdk clients

vinckr:
update Hydra SDK docs to the ory-client SKD
update Keto SDK docs to the ory-client SKD

TODO

- update SDK examples

--- 

IMO we can also merge this and update the examples after, it will be more work to test everything works.


docs/kratos/sdk/01_index.md
```diff
@@ -28,7 +34,7 @@ Code examples:
 
 - [Go](./05_go.mdx)
 
-Take a look at the source: [Generated SDKs for Ory](https://github.com/ory/sdk/tree/master/clients/client)
+Take a look at the source: [Generated SDKs for Ory]https://github.com/ory/sdk/tree/master/clients/client)
```
tomekpapiernik, line 37:
```suggestion
Take a look at the source: [Generated SDKs for Ory](https://github.com/ory/sdk/tree/master/clients/client)
```

docs/hydra/sdk/01_index.md
```diff
@@ -11,30 +11,36 @@ make backwards compatibility promises with regards to the generated SDKs. We hop
 Before you check out the SDKs, head over to the [REST API](reference/api.mdx) documentation which includes code samples for common
 programming languages for each REST endpoint.
 
+:::info
+
+`ory-hydra-client` SDK has been deprecated, please use `ory-client`!
```
tomekpapiernik, line 16: Do we have anything on `ory-client`? Are these SDKs officially deprecated?

docs/keto/sdk/01_index.md
```diff
@@ -13,17 +13,18 @@ documentation which includes code samples for common programming languages for e
...
-- [Python](https://pypi.org/project/ory-keto-client/)
-- [Ruby](https://rubygems.org/gems/ory-keto-client)
-- [Rust](https://crates.io/crates/ory-keto-client)
-- [Node.js REST](https://www.npmjs.com/package/@ory/keto-client) (with TypeScript)
-- [Node.js gRPC](https://www.npmjs.com/package/@ory/keto-grpc-client) (with TypeScript)
+- [Go REST](https://github.com/ory/client-go)
+- [Java](https://search.maven.org/artifact/sh.ory.keto/ory-client)
+- [Javascript REST](https://www.npmjs.com/package/@ory/client) with TypeScript definitions and compatible with: Node.js, React.js,
+- [Javascript gRPC](https://www.npmjs.com/package/@ory/keto-grpc-client) with TypeScript definitions and compatible with: Node.js,
+  React.js,
```
zepatrik, line 23:
```suggestion
- [Javascript gRPC](https://www.npmjs.com/package/@ory/keto-grpc-client) with TypeScript definitions and compatible with: Node.js
```
gRPC does not work in browsers

aeneasr:
Putting this on hold for a second, will need to figure out how to deal with self-deployed OSS /cc @zepatrik

docs/hydra/sdk/01_index.md
nbrugger-tgm: The `ory-client` sdks still exist but they are only for the versions <= `0.2.0` since nearly nobody uses this version anymore its imo more apropriate to link the most recent SDK. But no they are not deprecated if you use Hydra v0.2.0 they are perfectly valid

nbrugger-tgm:
Hi i came here because i did the exact opposite of this (just for kratos) i replaced ``ory-client`` with ``ory-kratos-client``. Because of the versioning and github activity i was sure that the "specific client" (eg. kratos-client, hydra-client) are the more recent ones and that the "general client" "go-client" is the deprecated one

nbrugger-tgm:
https://github.com/ory/docs/pull/1062

aeneasr:
We will definitely need to straighten this up. I'm thinking of maybe making explicit that the specific clients are more useful when self-hosting ory software, and the other client is useful when using the ory network. Maybe by naming it something like

```
ory-selfhosted-identity
```

but it's not really final yet...

nbrugger-tgm:
In this case i will close my PR since it would need a decition from ory side how the different clients will be handled/named

docs/hydra/sdk/01_index.md
vinckr: I think the notice is more confusing then helpful, I removed it. 
We should just make a proper announcement that you should move to the ory-client if possible.

zepatrik (approved):
I did not check all the links, but looks good AFAIK

vinckr:
fixing conflict now, then we can merge

PR #1119, opened 2022-12-01
feat: update data mapping section and remove obsolete sections

aeneasr:
Just brushing up social sign in. Mostly worked on data mapping and account linking.

Otherwise mostly removed stuff that doesn't apply to ory network


tomekpapiernik (changes requested):
Please don't merge before I push my edits. Thanks.

docs/kratos/social-signin/95_account-linking.mdx
```diff
@@ -0,0 +1,68 @@
+---
+id: account-linking
+title: Manage social sign-in providers linked to user accounts
+sidebar_label: Account linking
+slug: social-sign-in-account-link-connect-multiple
```
tomekpapiernik, line 5:
```suggestion
```

My recommendation: let's not use the `slug` front matter option. The content of the doc + title + id + h2/h3 content makes for good SEO. `Slug` just makes the front matter thicker w/o much benefits IMO. 

The required front matter content is codified in the guidelines. I'm aiming to make it standard across the entire documentation set. 

If you want to use `slug`, let's set up some rules on how to create them (length, word choice), and codify that in the guidelines first.

tomekpapiernik (changes requested):
Hey, @aeneasr I pushed my edits. Leaving you with one comment. I hope you'll agree with me. Cheers!

docs/kratos/social-signin/95_account-linking.mdx
aeneasr: The URL is the most important SEO signal. We discussed this during the planning. Let's use it!

> If you want to use slug, let's set up some rules on how to create them (length, word choice), and codify that in the guidelines first.

Rule of thumb is: Use the important keywords as part of the URL. What could be improved here is that `social-signin` is already in the path before, so this could be:

```suggestion
slug: link-multiple-provider-account
```

Adding all improtant keywords:

* ory
* account
* social sign in
* link
* (social) provider

etc

It's not an exact science, and will need refinement over time.

tomekpapiernik: Yes, I remember you mentioned this. However, quick search doesn't point to URLs being the most important. 

https://www.searchenginejournal.com/ranking-factors/urls/

> Evidence directly from Google indicates URLs do not play as great a role in search rankings as claims suggest.

> Looking back at the times Google has acknowledged URLs as a ranking factor, it seems the impact has waned over > time.

> In 2016, Google’s John Mueller [confirmed](https://www.youtube.com/watch?v=4FkSZIW6d48) keywords in a URL are a ranking factor.

> However, he described the signal as being “very small.”

> “I believe that’s a very small ranking factor, so it’s not something I’d really try to force. And it’s not something where > I’d say it’s even worth your effort to kind of restructure your site just so you can include keywords in the URL.”

https://seopressor.com/blog/url-structure-affect-seo/

image



Of course, the URL plays a role in SEO. But with the information I see on how much of a role it plays, I think that the document ID makes for a strong enough URL. Therefore, using `slug` isn't vital, which will help keep the front matter lean.

aeneasr: Interesting though that the URLs you found for the reasearch have the important keywords in the URL ;)

* https://seopressor.com/blog/url-structure-affect-seo/
* https://www.searchenginejournal.com/ranking-factors/urls/

Even if it has changed that the URL is not the most important factor, it is one of the factors that comprise seo.

Keeping the front-matter lean is a pure stylistic choice and personal preference. Trying to improve SEO is a practical guideline which helps Ory as a business. Ory will have to experiment with SEO. If it doesn't have an impact we can always change slugs or remove them iteratively. But not trying it because it makes the front matter look bloatet in markdown is not a argument that changes my opinion.

PR #1142, opened 2022-12-06
docs: explain pre and post persist registration webhook

sthh:
Explain pre and post persist registration webhook.

See https://github.com/ory/kratos/issues/2767

## Checklist

### DIFF
--- a/docs/kratos/hooks/01_configure-hooks.mdx
+++ b/docs/kratos/hooks/01_configure-hooks.mdx
@@ -414,6 +414,28 @@ Example payload:
 }
 ```
 
+When using an `after` registration hook, you can define the specific point in the flow where this webhook is called:
+
+- When the webhook can interrupt the flow (`can_interrupt: true`), the logic is called before the system creates a new identity.
+- When the webhook can't interrupt the flow (`can_interrupt: false`), the logic is called after the system creates a new identity.
+
+You can call a webhook before and after the system creates an identity. To do that, define two webhooks with different
+`can_interrupt` values:
+
+```yaml
+flows:
+  registration:
+    after:
+      password:
+        hooks:
+          - hook: web_hook
+            config:
+              can_interrupt: true
+          - hook: web_hook
+            config:
+              can_interrupt: false
+```
+
 :::tip
 
 Flow-interrupting webhooks work best if you give users meaningful information about the status of their flow. The best place to


docs/kratos/hooks/01_configure-hooks.mdx
```diff
@@ -414,6 +414,25 @@ Example payload:
...
+    after:
+      password:
+        hooks:
+          - hook: web_hook
+            config:
+              can_interrupt: true
+          - hook: web_hook
+            config:
+              can_interrupt: false
+```
```
tomekpapiernik, line 434:
```suggestion
When using an `after` registration hook, you can define the specific point in the flow where this webhook is called:

- When the webhook can interrupt the flow (`can_interrupt: true`), the logic is called before the system creates a new identity. 
- When the webhook can't interrupt the flow (`can_interrupt: false`), the logic is called after the system creates a new identity. 

You can call a webhook before and after the system creates an identity. To do that, define two webhooks with different `can_interrupt` values:

```yaml
flows:
  registration:
    after:
      password:
        hooks:
          - hook: web_hook
            config:
              can_interrupt: true
          - hook: web_hook
            config:
              can_interrupt: false
```
```

**Review by tomekpapiernik — CHANGES_REQUESTED** [review]
Well written, thanks! 
I tweaked the writing a bit more and added a list which makes everything better. Instantly. ;)
When you apply my suggestion please make sure to format the doc. again by running `make format`. Thanks!

File: docs/kratos/hooks/01_configure-hooks.mdx [hunk 2]
```diff
@@ -414,6 +414,25 @@ Example payload:
 }
 ```
 
+For the `after` registration hook, `can_interrupt` also defines the specific point in the flow where this webhook is called. If
+the webhook can interrupt the flow, it is called before the new identity is created. If the webhook cannot interrupt the flow it
+is called after the creation of the new identity. It is possible to call a webhook before and after creation of the identity by
+defining two webhooks:
```
tomekpapiernik, line 420:
```suggestion
When using an `after` registration hook, you can define the specific point in the flow where this webhook is called:

- When the webhook can interrupt the flow (`can_interrupt: true`), the logic is called before the system creates a new identity. 
- When the webhook can't interrupt the flow (`can_interrupt: false`), the logic is called after the system creates a new identity. 

You can call a webhook before and after the system creates an identity. To do that, define two webhooks with different `can_interrupt` values:
```

sthh:
Tanks Tom! I've applied your suggestions and rebased the PR.

PR #1150, opened 2022-12-09
docs: adding documentation for advanced settings

kelkarajay:
Documentation related to the feature - https://github.com/ory-corp/cloud/pull/3414

## Related issue(s)

https://github.com/ory-corp/cloud/issues/3283

### Via Console 
Screenshot 2022-12-16 at 11 51 16

### Via CLI
Screenshot 2022-12-16 at 11 52 23


docs/advanced/settings.mdx
```diff
@@ -0,0 +1,28 @@
...
+after the first query. The tradeoff worth considering is that the embedded identity representation will have eventual consistency
+(post expiry of the cached session).
+
+:::note
+
+Projects on Business plan or higher can use the Session caching feature.
+
+:::
+
+Once the feature has been turned on, the cache duration can be controlled by the clients by setting the `max-age` value in the
```
CaptainStandby, line 26: Maybe you could add an example how this would be implemented?

tomekpapiernik:
Hey @kelkarajay, this doesn't really fit the structure we're working toward right now. Did you talk to @aeneasr to establish where this topic belongs?

aeneasr:
There's a page for this already: https://www.ory.sh/docs/concepts/cache

Maybe just adding how to enable the feature there is enough?

docs/advanced/settings.mdx
kelkarajay: yep. Moved and added configuration steps in  https://www.ory.sh/docs/concepts/cache

kelkarajay:
@aeneasr @tomekpapiernik @CaptainStandby I've done changes. Take a look at the updated screenshots in the PR description

vinckr (approved):
lgtm

CaptainStandby:
looks good

CaptainStandby (dismissed):

docs/concepts/cache.mdx
```diff
@@ -66,3 +71,52 @@ ory.toSession(undefined, undefined, {
...
+ory list projects
+
+## Get config
+ory get identity-config {project-id} --format yaml > identity-config.yaml
+```
+
+2. Add the configuration for session caching
+
+```yaml title="config.yml"
+feature_flags:
```
aeneasr, line 109: Does this actually work? I remember seeing somewhere that we disabled this?

tomekpapiernik:
In general, the way this document is put together leaves me a little confused. Reading from the top, I'd assume that session caching just works, but then, at the very end of the doc, there's the configuration section. If I understand this situation correctly, we should say that this feature is not standard, not active by default, and that it is available only for Business and Enterprise plans. Then explain what it does and how to enable it.

docs/concepts/cache.mdx
tomekpapiernik: I just tried this - pops up when you get the Identities config, throws an error when I tried to set to `true` on a non-business plan. 

```
The request was malformed or contained invalid parameters
This feature is not available on your subscription plan.
```

kelkarajay: Yes, it does. There's a validation to ensure only business plans or higher can do this

tomekpapiernik:
@kelkarajay can you help me understand the situation and answer the question I asked in my previous comment? Thanks :)

kelkarajay:
@tomekpapiernik Sorry, I misunderstood. I thought the question was for @aeneasr from a product standpoint. From what I see, we do have a higher plan (an Enterprise offering?) than the Business plan. 
Screenshot 2022-12-21 at 15 06 36

-> **So I'd keep it "Business plan and higher".** 

As far as the other question, about the feature restriction note placement - I can move the note right under the feature header (but imo this doesn't solve anything), since the document initially promises a low latency global network unless you use the highest tiers. Thoughts on that placement?

Screenshot 2022-12-21 at 15 16 48

aeneasr:
Hi, yes this feature is currently only business and enterprise

tomekpapiernik (approved):

kelkarajay:
looks good to me

PR #1163, opened 2022-12-16
docs: update native-browser.mdx

Maragues:


vinckr (approved):
Thanks for catching that! 🙌

PR #1188, opened 2023-01-02
feat: generated index cards for main categories

vinckr:
What do you think of using these generated index pages? 
you get to them when you click on one of the major categories. 
in prod we would have to tweak it a bit to look nice, this is just to give a tangible idea of what i am talking about 

Screenshot 2023-01-02 at 19 19 33

## Checklist


aeneasr:
I think we shoudl try it, because before it was not possible to link to these pages. Now it seems to be possible :)

https://docs-2oqyuyz13-ory.vercel.app/docs/category/identities-and-sessions

aeneasr (approved):

tomekpapiernik (changes requested):
I don't like the way this looks and how it affects the UX. First of all, to me it just looks ugly and without proper styling it downgrades the "docs experience" compared to the recent small upgrade that was caused by slight theme tweaks around code blocks. 

Additionally, IMO it impacts the UX negatively. Currently, if you're reading document A and then a question/idea pops into your head, you are free to expand the main categories, browse the left-hand navigation and still have the document you were reading open. 
When we add the cards, any click on one of the main categories takes you away from the document you were reading.

vinckr:
I hear you @tomekpapiernik , just a small rebuttal on some of your points.

Btw we can also link directly to a document for a category instead of the cards. 
Even we choose to not use the cards at all, I would still link each category to one document when you open them.

> then a question/idea pops into your head, you are free to expand the main categories, browse the left-hand navigation and still have the document you were reading open.

I attached a small video that shows that this is still the case with the index cards. If you click on the arrow the menu pops up, pretty straightforward IMO.

https://user-images.githubusercontent.com/55394283/210324517-4f9a0fad-46e9-44a3-be7b-464eda7dfc45.mov

As for the design,
we would need to add a "starting sentence" to most documents, that summarizes the document in one sentence. I think that would be a useful addition in any case, rarely do I want to read a document and find out in the end if it is useful.

Was there anything else particular in the design that stood out to you? Apart from the text the boxes look fine to me.

Generally this PR would have these benefits IMO:
- we can link to a category
- we have control what the user sees when they open a category
- we can structure categories better (having a similar flow for every category long term)

I personally find the cards also more easily scannable and structured as opposed to the sidebar view when I come into a category as total newcomer. 

Phew wall of text, but it hopefully makes clear why I think this would be a useful addition. 

I cant improve the design, but if you think it would make this mergeable, I am happy to give this PR some polish (i.e. make the card descriptions concise for each document and choose a good document to link to where the cards are not useful).

tomekpapiernik:
Reg. the video example you posted - it's cool that it works, I didn't notice that at first. However, the most natural thing to do is to simply click the category name in the sidebar rather than precision aim at the arrow used for expanding, don't you think?  If you click the category name instead of the arrow, the behavior I described is true. 

As for the way this looks in Docusaurus, I think that it currently just looks bland and very basic. The images used in buttons (?) don't match our color theme and are kinda hard to identify (I had to zoom in like 200% to see that one of the icons used is a filing cabinet drawer). 

Also, I don't think Docusaurus lets you add descriptions for groups of docs. Single docs? Sure. Have a look at their own example for using this feature: https://docusaurus.io/docs/category/guides
If this is true, lone docs like the OAuth2 "Introduction" get a description in the generated "tile", but groups like "Debug and Help" don't. :(

PR #1219, opened 2023-01-18
feat: enable syntax highlighting for relationtuples and relationships

zepatrik:
This adds a new language to represent the new relationship string format, as well as re-enables syntax highlighting for the current relation tuples.

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,9 @@ slug: excellent-feature-for-some-reason
   ```
   ````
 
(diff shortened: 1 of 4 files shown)


vinckr (approved):
I checked a few pages and LGTM.

Could you maybe add some documentation in the README how to use it, if it is relevant to documenters 🙏

tomekpapiernik:
The trigger is `keto-relation-tuples`, right @zepatrik ?
If that's correct, I'll mention it in the guidelines.

zepatrik (commented):
> The trigger is keto-relation-tuples, right @zepatrik ?

yes exactly, I'll add that to the readme

PR #1232, opened 2023-01-26
docs: explain csrf defenses

vinckr:
Adds explanation to the security / threat model document on how Ory Identities handles CSRF attacks.

### DIFF
--- a/docs/kratos/concepts/security.mdx
+++ b/docs/kratos/concepts/security.mdx
@@ -30,6 +30,26 @@ for example login and registration endpoints.
 When self-hosting the Ory Kratos Identity Server, it's the responsibility of the administrator to implement and manage rate
 limiting or other measures to ensure the security of the network.
 
+## Defenses against cross-site request forgery
+
+Cross-site request forgery, also known as CSRF or XSRF, is a type of attack that occurs when a malicious website or attacker
+tricks a user's web browser into sending a request to another website without the user's knowledge or consent. In the context of
+Ory Identities, it is a security vulnerability that can be exploited by an attacker to gain unauthorized access to a user's
+account or perform actions on their behalf.
+
+For example, imagine an attacker creates a website that has a form that, when submitted, sends a request to Ory Identities to
+change a user's password. If a user is logged in and visits the attacker's website, their browser may automatically include their
+Ory session cookie in the request, allowing the attacker to change the user's password without their knowledge or consent. To
+prevent this Ory Identities uses a cookie-based security model with the [`sameSite`](samesite at w3) attribute on the cookie. This
+means that the Ory session cookie can only sent with requests that originate from the same website or origin that set the cookie,
+effectively protecting against cross-site request forgery (CSRF) attacks.
+
+This is a security feature that allows Ory Identities to only recognize requests made by the same website, preventing a malicious
+website from making a request to Ory Identities on behalf of the user. This way, even if an attacker is able to create a request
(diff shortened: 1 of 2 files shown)


docs/kratos/concepts/security.mdx
```diff
@@ -30,6 +30,26 @@ for example login and registration endpoints.
...
+Ory session cookie in the request, allowing the attacker to change the user's password without their knowledge or consent. To
+prevent this Ory Identities uses a cookie-based security model with the [`sameSite`](samesite at w3) attribute on the cookie. This
+means that the Ory session cookie can only sent with requests that originate from the same website or origin that set the cookie,
+effectively protecting against cross-site request forgery (CSRF) attacks.
+
+This is a security feature that allows Ory Identities to only recognize requests made by the same website, preventing a malicious
+website from making a request to Ory Identities on behalf of the user. This way, even if an attacker is able to create a request
+that includes a user's session cookie, the request will not be accepted by Ory Identities as it will not be made from the same
+website. It ensures that only requests from the same origin are able to use the cookie, providing an additional layer of security
+against CSRF attacks.
```
aeneasr, line 51: This isn't correct unfortunately. CSRF can happen without a session too. One such attack is login csrf ( https://support.detectify.com/support/solutions/articles/48001048951-login-csrf )
. We deploy regular CSRF counter measures as explained here:

- https://owasp.org/www-community/attacks/csrf
- https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

Instead of making the session cookie CSRF resistant, we have a dedicated CSRF cookie whose purpose it is to protect against CSRF, even if no session exists yet.

aeneasr (changes requested):
See comment

docs/kratos/concepts/security.mdx
vinckr: > Instead of making the session cookie CSRF resistant

We still use the [sameSite attribute](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#samesite-cookie-attribute) as well though, right? 

Is the dedicated CSRF cookie what they describe in "[Double Submit](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie)" cookie?
or what they describe under Login CSRF?

> Login CSRF can be mitigated by creating pre-sessions (sessions before a user is authenticated) and including tokens in login form. 

If it is really complex to describe here, maybe it is not worth it to go in such detail and just say there are multiple protections against multiple threat vectors regarding CSRF?

aeneasr: No we use https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern

vinckr:
Thanks for the clarification!


- Rewrote the CSRF text to clarify that we mainly use the anti-csrf cookie
- Removed the "What is CSRF" section from the troubleshooting doc and moved that content also to the Threat models doc

This way you can learn about CSRF and how Ory protects against it in the threat models doc and the troubleshooting is just focused on how to fix errors related to that.

Feel free to take another look :)

fbolton (approved):
Text and layout looks good to me.

vinckr:
Can you please take a quick look before we merge @aeneasr ?

aeneasr (approved):
Great!

PR #1296, opened 2023-03-03
fix: update SDK examples

### LINKED ISSUE #1223: Clarify SDK warning
We should make clearer which SDK you use in what instance, so the SDK listed at 
https://www.ory.sh/docs/sdk for Ory Network 
and the SDKs in the self-hosting reference docs for each service
currently here https://www.ory.sh/docs/ecosystem/projects for self-hosting Ory services

From [slack](https://ory-community.slack.com/archives/C012RJ2MQ1H/p1674225530842849?thread_ts=1674156770.593119&cid=C012RJ2MQ1H)
> I would recommend more than a warning! There are or can be differences that are breaking. Things almost working is much worse than not working at all. Maybe even put in a check when using various APIs/SDKs that will fail out of hand if mixing Ory Network/Self-hosted servers and APIs.

---
Current admonition:

INFO
The client SDK works with self-hosted Ory services, but it is recommended to use the dedicated client. Read more in the [self-hosting documentation](https://www.ory.sh/docs/ecosystem/projects) for each project under Reference/SDK.

TODO:

- change to WARN and update text to be more clear that you should definitely use the other SDK when self-hosting.

request for comment: @tomekpapiernik @fbolton

vinckr:
- update the Kratos Go SDK example to work with Ory Network playground
- create a dedicated Go SDK example document for Ory Kratos self-hosted
- add section for self-hosted examples in sdk sidebar
- change main sdk page to clarify sdk warning

closes #1223

Still in draft stage, bc it only updates the Kratos Go SDK example. I would add Hydra + Keto next. 

@tomekpapiernik Do you think this is a good way to structure the examples? I would add a selfhosted folder for hydra and keto then as well. suggestions welcome 🤗 

@aeneasr What do you think of the examples split up like this? The difference is mostly in the steps to run it and the basepath I guess, but this would help prevent people from getting confused with kratos-client / ory-client IMO.

## Checklist


tomekpapiernik:
IMO there should be just categories for Ory Network / OS Servers, so: 

- Ory Network
- Ory Kratos
- Ory Hydra
- Ory Keto

These categories would then have an intro and examples for specific programming languages. 
For Ory Network, on the top level, I'd like to see examples divided into different programming languages. Within the "Go" doc I'd like to see examples of Identities / OAuth2 / Permissions operations simply organized by H2s.

vinckr:
TODO

- [ ]
- - [ ]
- - [ ]
- - [ ]

vinckr:
I tested all the Ory Network SDK examples, as they are very easy to test. 
The self-hosted I tested a bit, but not completely, much harder to set up. 
Maybe a generic docker-compose that sets all services up and runs tests in the future....

PR #1305, opened 2023-03-08
Remove deprecated --id option

sezanzeb:
Setting the client id is not possible anymore (https://github.com/ory/hydra/issues/2911), but the --id option was still in the docs.

## Checklist

I'm not going to read through pages of legal lingo and am not going to sign anything to remove a deprecated line of documentation. Replicate the change with your own PR if you can't merge it like that.


vinckr:
Hello @sezanzeb 

thanks!  🙌 
We need to update the source of these CLI docs, otherwise this gets overwritten once they are generated anew.

If you are open to do that, here is some documentation how to find the files you need to edit: https://github.com/ory/docs#cli-and-api-reference---auto-generated-content

vinckr:
The CLI doc has since been updated, 
thanks for the report anyway @sezanzeb 
Let us know if you find something else that is not right!

PR #1312, opened 2023-03-13
fix: add missing doc to sidebar

aeneasr:
## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [ ] I have referenced an issue containing the design document if my change introduces a new feature.
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security vulnerability. 
      If this pull request addresses a security vulnerability, 
      I confirm that I got approval (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added the necessary documentation within the code base (if appropriate).

PR #1313, opened 2023-03-13
docs: self-host production intro

vinckr:
as discussed here, this removes the scary admonition with something more helpful, and streamlines it across the projects: https://github.com/ory/docs/pull/1312


fbolton (approved):
LGTM!

PR #1316, opened 2023-03-14
fix: metadata fields in actions docs

kmherrmann:


kmherrmann:
@CaptainStandby please check that i'm not wrong here, but the docs seem inconsistent (and a customer just stumbled on it)

aeneasr (approved):

PR #1347, opened 2023-03-29
fix: add oathkeeper obs port

vinckr:
Forgot to add the Oathkeeper port in my previous PR. 
Also words the sentence less confusing, they all have the same endpoint but different ports in the default configuration.


fbolton (approved):
LGTM!

PR #1364, opened 2023-04-12
docs: make Ory logo link to website, and remove obsolete Website link

fbolton:
Implements the universal convention whereby the company logo in the top right of the page brings you to the main corporate web site. This also makes the "Website" link on the navbar redundant.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -78,7 +78,7 @@ const config = {
         alt: "Ory",
         src: `/docs/img/logo-docs.svg`,
         srcDark: `/docs/img/logo-docs-dark.svg`,
-        href: `https://www.ory.sh/docs`,
+        href: `https://www.ory.sh`,
       },
       items: [
         {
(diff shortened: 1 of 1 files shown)


vinckr (approved):
good change!

tomekpapiernik (approved):
image

jonas-jonas (commented):
My two cents: I hate when docs do this. Most of the time, I want to go to the root of the docs, instead of the landing page. As getting to the root of the docs is pretty annoying if there is no link to it on the top left of the page. Any particular reason for this change?

tomekpapiernik:
@jonas-jonas There's the "Documentation" link in the top bar and it does exactly that - takes you to the "root" which, in our case, is the "welcome" page. ;)

fbolton:
@jonas-jonas The reason for this change is that it's a universal convention on the Internet, so most users would have exactly the opposite expectation. They would click on the company logo expecting to get back to the main site and be surprised when they stay on the docs page.

And as Tom says, you can always click on the Documentation tab to get back to the welcome page.

PR #1373, opened 2023-04-20
fix: split up custom ui examples

vinckr:
- creates separate page and sidebar item for custom ui examples/quickstarts
- adds a mention of the custom ui examples on next steps of custom UI overview.

tangentially related minor fixes:
- removes outdated supabase example from overview
- minor style improvements

We can probably link to the custom ui examples in other places as well, but this felt like the most natural start - I did not want to go through all docs and see where it could fit.


fbolton (approved):
LGTM!

PR #1402, opened 2023-05-12
fix: self-hosted social signin docs

vinckr:
- fix some broken links in the social signin doc for selfhosted
- add example config 
- rephrase some things to be less confusing.


kmherrmann (commented):
lgtm!

PR #1462, opened 2023-07-18
docs: update webhooks ips

adamwalach:
## Checklist

### DIFF
--- a/docs/guides/integrate-with-ory-cloud-through-webhooks.mdx
+++ b/docs/guides/integrate-with-ory-cloud-through-webhooks.mdx
@@ -48,6 +48,7 @@ these specific IPs:
 - 35.242.228.133
 
 These addresses will be used in the US region:
+
 - 35.245.69.134
 - 34.94.111.83
 
--- a/docs/guides/integrate-with-ory-cloud-through-webhooks.mdx
+++ b/docs/guides/integrate-with-ory-cloud-through-webhooks.mdx
@@ -46,6 +46,8 @@ these specific IPs:
 
 - 34.22.170.75
 - 35.242.228.133
+- 35.245.69.134
+- 34.94.111.83
 
 By setting your firewall rules to allow traffic only from these IP addresses, you'll be adding an extra layer of protection
 against any unwarranted access attempts.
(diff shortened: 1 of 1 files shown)


kmherrmann:
Do we need to communicate this to all our clients?

vinckr:
Would be good to include in status.ory.sh as maintenance notice as well as a short announcement in the Slack - we can also add a note to the newsletter. 

I am not sure what the impact is - is this relevant if you have some firewall rules around webhooks from Ory Network?

adamwalach:
Yes, I will create a notice in status.ory.sh and send a slack message, the same as we did last time: https://status.ory.sh/maintenance/212371

vinckr:
> Yes, I will create a notice in status.ory.sh and send a slack message, the same as we did last time: https://status.ory.sh/maintenance/212371

perfect, I think we can document that as standard practice whenever we make these types of changes that might affect some customers in production.

kmherrmann:
Maybe we should also send an email about this if it has the potential to break webhooks for those with firewalls?

vinckr:
if you are subscribed to status.ory.sh (which I would recommend to everyone in production) you would get an email I think. 
But yea we can include it in the newsletter tomorrow? Or is there a "maintenance" email channel/audience as well?

kmherrmann:
I think we should create this audience with everyone who is an ory network project member on it.
We can restrict to paid projects to reduce noise.

kmherrmann:
Newsletter is a good idea but also something that’s easy to overlook

vinckr:
Good point. Something like the security newsletter but just for Ory Network? 
Maybe not the right place to discuss it here, but I also need this audience to do the support stuff on Slack. So would go hand in hand to do this in Hubspot i guess.

kmherrmann:
We should have an email list with all Ory Network users (which are members of a project). We can consider only including paid projects or somehow cutting noise by pruning projects that haven't seen activity in 90 days.

This list should not need an opt-out and may only be used for maintenance notifications and potentially sharing information like post mortems. absolutely no marketing, purely necessary information that we expect admins to read. low volume.

aeneasr:
us is now available - merge this?

kmherrmann (approved):

vinckr (approved):

PR #1485, opened 2023-08-03
fix: update basePath config in index.js

OryTiki:
AFAI understood, the basePath in the SDK configuration should point to the proxy and not to <slug>.projects.oryapis.com. When pointing to oryapis.com or the ORY_SDK_URL, the login works, but calling `createBrowserLogoutFlow` returns a logout url, which leads to an error (Unable to log out because the logout token in the URL query does not match the session cookie). 
TBH I'm not sure if the basePath should point to `http://localhost:4000/.ory` or `http://localhost:4000`

### DIFF
--- a/code-examples/protect-page-login/expressjs/routes/index.js
+++ b/code-examples/protect-page-login/expressjs/routes/index.js
@@ -9,7 +9,7 @@ var sdk = require("@ory/client")
 var ory = new sdk.FrontendApi(
   new sdk.Configuration({
     basePath:
-      process.env.ORY_SDK_URL || "https://playground.projects.oryapis.com",
+      "http://localhost:4000/.ory",
   }),
 )
 // highlight-end


code-examples/protect-page-login/expressjs/routes/index.js
```diff
@@ -9,7 +9,7 @@ var sdk = require("@ory/client")
 var ory = new sdk.FrontendApi(
   new sdk.Configuration({
     basePath:
-      process.env.ORY_SDK_URL || "https://playground.projects.oryapis.com",
+      "http://localhost:4000/.ory",
```
aeneasr, line 12: This API call doesn't need to go though the proxy - so it's easier / better / more flexible if it just points to the regular SDK url. I think this was fine the way it was, or did you encounter any issues?

aeneasr (changes requested):

code-examples/protect-page-login/expressjs/routes/index.js
OryTiki: I did encounter an issue (on localhost) when trying to implement the logout method via `createBrowserLogoutFlow`. Redirecting the user to the provided  `logout_url`, led to the error: Unable to log out because the logout token in the URL query does not match the session cookie.
My fix was to set the basePath to localhost:4000. After that change, the logout_url is working as desired

Benehiko: maybe the guide is inconsistent with the example code. or it has been changed.
I don't think this example uses the integrations package to proxy requests to itself and thus you need to spin up the proxy / tunnel externally. 

Using the upstream Ory APIs would thus give infinite redirect loops since the cookie is not set correctly. So i think this change is actually correct.

aeneasr: Why is the CI then failing? :D

The set up for this app is `ory proxy -> node app` iirc. The node app calling the ory APIs will not send any cookies anyways. I don't think this makes sense.

Will need to look into the reason why logout_url isn't working though

Benehiko: i'll take a look

vinckr:
closing this but should be investigaed

aeneasr:
If it needs investigation please create an issue or keep this open!

PR #1499, opened 2023-08-10
feat: add CTA to ory network

LBBO:
This CTA differs a bit from the design in the interest of saving time.


aeneasr:
Isn't this missing some button i can click on?

LBBO:
Fixed. Apparently the docusaurus production build just deletes `@container` queries, lol. But there ended up being no necessity for them, so I just exchanged them for equivalent media queries.

OryTiki (changes requested):
Still need some work on the button styling or content to ensure most viewports look good. 

image

src/components/OryNetworkCta/ory-network-cta.css
```diff
@@ -0,0 +1,134 @@
...
+  text-decoration: none;
+}
+
+.ory-network-cta__inline-get-started {
+  display: block;
+}
+
+@media (min-width: 1070px) {
+  .ory-network-cta__button {
+    display: block;
```
jfcurran, line 94:
```suggestion
    display: block;
    text-align: center;
```

jfcurran (changes requested):
the button text should be centered, this change will do the trick

vinckr (commented):
lgtm. 

only nit is that its not present in the "self-hosting" category, is that on purpose?

aeneasr:
@vinckr it's only visible if there is a table of contents on the right side: https://docs-fgqd0n2bm-ory.vercel.app/docs/ecosystem/configuring

PR #1555, opened 2023-09-29
feat: update b2b sso organization documentation

jonas-jonas:
## Checklist

### DIFF
--- a/docs/kratos/organizations/organizations.mdx
+++ b/docs/kratos/organizations/organizations.mdx
@@ -33,84 +33,14 @@ graph LR
 
 ## Manage organizations
 
-Organizations can be created, updated, and deleted via the HTTP REST API. The following examples show how to manage organizations.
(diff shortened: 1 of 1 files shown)


vinckr (commented):
Its live now right @jonas-jonas ?

Btw do you think it makes sense to keep the example how to configure it via API in, or was that more hacky than it should be?

jonas-jonas:
Ah, yes, it is. 

The API example is really hacky, and we shouldn't have that in the documentation, IMO. The CLI will support it soon (ish): https://github.com/ory/cli/pull/318

But since it doesn't support it yet, might be good to keep the API example in there... I guess I'll revert the removal for now.

vinckr:
> I guess I'll revert the removal for now.

Your call 😁 
IMO if there are concrete plans to add it to the CLI, we can remove it. 
I dont see too many users using it in that way tbh. 
But if the CLI/API plans are still a longer time out (say Q1'24) then I would leave it in.

vinckr (approved):

PR #1578, opened 2023-10-13
fix: clarify pagination

alnr:


kmherrmann (approved):

vinckr (commented):
👍 

Do we also need to change things in the [SDK docs](https://github.com/ory/sdk/blob/master/clients/client/go/docs/IdentityApi.md#listidentities)?

alnr:
> 👍
> 
> Do we also need to change things in the [SDK docs](https://github.com/ory/sdk/blob/master/clients/client/go/docs/IdentityApi.md#listidentities)?

Yes, those need updating.

aeneasr:
Those should update themselves automatically. However, these do not: https://www.ory.sh/docs/ecosystem/api-design#pagination

Issue #1213, opened 2023-01-13
Request flow diagrams not very visible in dark mode


raymond-design:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/docs/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/docs/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Most of the request flow diagrams in the documentation is not very visible in dark mode because the text is almost the same color as the background.

### Describe your ideal solution

We can simply change the color theme a little bit and the diagrams will be much more visible.

### Workarounds or alternatives

n/a

### Version

Newest Ver. of the Docs

fbolton:
Not a quick fix, I'm afraid, but we'll look into it.

PR #1627, opened 2024-01-15
fix: add dark mode for diagrams (#1213)

iverly:
This pull request adds the dark mode theme for mermaid diagrams that match the selected theme.

#1213 

## Checklist

## Further comments

This fix only applies after a page refresh if the page has a diagram. This is due to the mermaid package refreshing it empty (and I don't know why).

We can open a new issue to fix this bug later, while still keeping diagrams more visible than today.

### DIFF
--- a/src/theme/Mermaid.js
+++ b/src/theme/Mermaid.js
@@ -13,6 +13,7 @@
 // `}/>
 
 import React, { useEffect, useState } from "react"
+import { useColorMode } from "@docusaurus/theme-common"
 import mermaid from "mermaid"
 import styles from "./mermaid.module.css"
 import cn from "classnames"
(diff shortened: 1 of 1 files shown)


vinckr (approved):
Thanks a lot for the contribution! 
This fixes our issues with darkmode diagrams :tada: 

It works great when I have dark mode enabled on the docs and then navigate to a page that has a mermaid diagram.
One thing that is still not fixed it seems:
If I am already on the page (for example [this one](https://docs-92f90e7xq-ory.vercel.app/docs/kratos/self-service/flows/user-registration)) and then toggle dark mode, its the old unfixed version. A page refresh F5 fixes the diagram, but that might not be obvious to users) 

Still this is a 90% fix IMO, thanks again :+1:

src/theme/Mermaid.js
```diff
@@ -43,11 +44,19 @@ const Mermaid = ({ chart }) => {
...
+    mermaid.render(
+      id,
+      `%%{init: {'theme':'${
+        colorMode === "light" ? "neutral" : "dark"
+      }'}}%%\n${chart}`,
+      (svg) => {
+        setSvg(svg)
+      },
+    )
   }, [])
```
zepatrik, line 60: It should be enough to add the color mode here to not require reloads.
```suggestion
  }, [colorMode])
```

zepatrik: Ok, so let's merge as is and we'll figure out the rest later. It is a huge improvement anyway, as @vinckr said. Thanks again! :tada:

PR #1636, opened 2024-01-31
fix: typo in Hydra docs

LucianBuzzo:


vinckr (approved):
thanks for catching this!

PR #1656, opened 2024-02-27
feat: add STS external-id to event stream trust policy

alnr:


docs/actions/live-events.mdx
```diff
@@ -52,18 +52,24 @@ Record the ARN of the IAM role you created, for example:
...
       "Effect": "Allow",
       "Principal": {
-        "AWS": "arn:aws:iam::601538168777:user/event-streamer"
+        "AWS": "601538168777"
       },
-      "Action": "sts:AssumeRole"
+      "Action": "sts:AssumeRole",
+      "Condition": {
+        "StringEquals": {
+          "sts:ExternalId": "b21b757b-7d8e-4864-bb7b-800f5fad7aa6"
```
jonas-jonas, line 70: Maybe we should just use an obvious placeholder, so user's that just skim the page are not confused, when they paste it.

alnr: Valid consideration. The other example snippets on this page also use valid IDs instead of placeholders, so I did that here as well.

PR #1727, opened 2024-04-26
chore: fix commands

vinckr:
changing {...} to $.... in commands so you can copy & paste them

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,11 +238,12 @@ prevents leaking of sensitive data such as tokens or API keys.
 
 Follow these rules when using placeholders and dummy data:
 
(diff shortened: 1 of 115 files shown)


jonas-jonas (dismissed):
Nice, I think this is the right direction.

I'd prefer `$ORY_PROJECT_ID` as it clearly states what it does, and env variables are typically upper case. 
This way, someone using the docs can execute an `export ORY_PROJECT_ID ="uuid"` before running any of the commands, and it just works.

vinckr:
capitalized the commands now @jonas-jonas :)

zepatrik (approved):

PR #1757, opened 2024-06-13
chore: typo in go login example 

vinckr:
I also get a redirect loop when running the example, but not sure if that is my issue here or not 🤔 

## Checklist


vinckr:
Not sure why the test fails...

I get this error when running locally:
```
./middleware.go:45:30: app.ory.FrontendApi undefined (type *client.APIClient has no field or method FrontendApi)
```

aeneasr:
Probably a SDK version mismatch. Maybe you need to update the go.mod file

vinckr:
ah that was it! how can we prevent this in the future? 

IMO this is kinda bad since if we cant rely on the tests to tell us when something breaks, why have them in the first place? In this case the test even gives a false positive. 

(run `go get -u github.com/ory/client-go` during the tests or otherwise)

aeneasr:
I don't think this was a false positive, the code did not compile so the example was broken. So the test ensured we don't merge something that doesn't work to master. In my view, there is no need for change

vinckr:
> I don't think this was a false positive, the code did not compile so the example was broken. So the test ensured we don't merge something that doesn't work to master. In my view, there is no need for change

Yes, but the test did not catch this previously, we only noticed because we manually ran the example. 
So the test was all green the last weeks, but the example was actually broken all this time. 

Thus the test did not serve its purpose (to alert us if an example is broken) - or do I miss anything?

vinckr:
The issue is that the test has a pinned version of client-go and the example tells you to use the latest version. 
Its usually no problem, but if there are any breaking changes we wont notice AFAICT.

zepatrik:
There is no way to not pin the version, so the example always has to be based off of some version.
We could have some kind of auto-update automation that would then cause the tests to fail once there are breaking changes (which are something we always try to avoid though).

vinckr:
Makes sense. I guess its not super critical and we just look at the examples on the rare occasion that we have a breaking change.

PR #1766, opened 2024-06-19
fix: outdated keys for http courier templates

zepatrik:
closes https://github.com/ory/kratos/issues/3815

### DIFF
--- a/docs/kratos/self-hosted/03_mail-courier-http.mdx
+++ b/docs/kratos/self-hosted/03_mail-courier-http.mdx
@@ -71,7 +71,7 @@ available through a `ctx` object. `recipient` will always be the email address o
   - verification_code
   - identity
 
-A universal JSONNet template that works with all flows, would look like this:
+A universal Jsonnet template that works with all flows, would look like this:
 
 ```jsonnet
 function(ctx) {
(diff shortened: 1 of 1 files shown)


matdehaast:
@zepatrik not sure if its possible. But may be worth updating the changelog to notify of the breaking change.

docs/kratos/self-hosted/03_mail-courier-http.mdx
```diff
@@ -42,36 +42,36 @@ The configuration consists of:
...
 - verification_code_valid
-  - To
-  - VerificationCode
-  - Identity
+  - to
+  - verification_code
+  - identity
 
-A universal Jsonnet template that works with all flows, would look like this:
+A universal JSONNet template that works with all flows, would look like this:
```
alnr, line 74: > The name Jsonnet is a portmanteau of JSON and [sonnet](https://en.wikipedia.org/wiki/Sonnet), pronounced "jay sonnet".

from https://jsonnet.org/
:)

alnr (approved):

docs/kratos/self-hosted/03_mail-courier-http.mdx
aeneasr:
```suggestion
A universal Jsonnet template that works with all flows, would look like this:
```

PR #1795, opened 2024-07-17
feat: guides and docs for CLI v1

zepatrik:


vinckr (approved):
looks good to me 😁

aeneasr:
We still need to update all guides and tutorials with the new CLI usage right? Since it's already released on homebrew

zepatrik:
Yes, I'll open a separate PR for that. The CLI reference docs were also already updated, even before the first release.

PR #1800, opened 2024-07-22
Update 30_apple.mdx

mhaagens:
Add a note about the provider id needing to be "apple" because of CSRF middleware exemption.


vinckr (approved):
thx for the contribution 🎉 
i changed the text a bit and moved it to the other admonission

PR #1861, opened 2024-09-04
chore: mention more token+cookie prefixes

alnr:


jonas-jonas (approved):

docs/security-compliance/token-formats.mdx
```diff
@@ -25,13 +25,19 @@ security breaches and unauthorized access.
...
 Developers can configure the expiration time and cookie options for Ory session cookies to fit their application's specific needs.
 
-In summary, Ory session cookies are prefixed with ory*session* and are essential for maintaining user sessions. They can be
-customized to fit specific application needs and are secure and tamper-proof.
+## Other Ory cookies
+
+Ory Network will also issue other cookies to protect against CSRF attacks, provide load balancing, and other security features.
+These cookies will be prefixed as follows:
+
+- `ory_`: Ory-internal cookie (CSRF protection etc)
```
vinckr, line 38:
```suggestion
- `ory_`: Ory-internal cookie (CSRF protection)
```

IMO either list explicitly or not (I assume its not that much)

jonas-jonas (approved):
would be cool if we could dynamically create the full cookie name from the project slug, but that will require some changes around getting the current project into docs.

vinckr (approved):
lgtm, small nit

docs/security-compliance/token-formats.mdx
alnr: We have a bunch of other cookies which I'd only list of someone really, really, really needs to know their names for some inexplicable reason. GDPR certainly doesn't require listing them all by name.

PR #1887, opened 2024-09-23
docs: include more items on the sidebar

aeneasr:
## Checklist


docs/guides/authentication.mdx
```diff
@@ -0,0 +1,11 @@
+---
+id: authentication
+title: Authentication & Login
+---
+
+```mdx-code-block
+import DocCardList from '@theme/DocCardList';
+import {useCurrentSidebarCategory} from '@docusaurus/theme-common';
+
+<DocCardList items={useCurrentSidebarCategory().items}/>
```
aeneasr, line 10: @vinckr this is an easy way to create overview pages. It's obviously better to not auto-generate them, but it helps with one issue:

If the `sidebar_label` is set in a document, it overrides the sidebar label defined in the config. So while the sidebar config says the label is `foo`, if the label in the doc is `bar`, then the label is `bar`. This is problematic for categories with a link:

```
    {
      type: "category",
      label: "Identity management",
      collapsed: false,
      collapsible: false,
      link: {
        type: "doc",
        id: "kratos/manage-identities/overview", // if this has `sidebar_label: foo`, then the category will be called "foo"
      },
```

For this reason, these auto-generated categories pages work as fillers.

src/sidebar.ts
```diff
@@ -2,72 +2,79 @@
...
+              },
+            ],
+          },
+        ],
+      },
+    ],
+  },
+]
+
+const quickstartSidebar = (flat: boolean): ExtendSidebar => {
```
aeneasr, line 70: The idea is the following: the main documentation page shows the first 2 levels of the navigation (aka flat = true). If we browse on the main page (level 1), we see level 1 and 2 (intro, quickstarts, guides, ...). when we jump into one of the sub-items of a category (e.g. quickstarts), we jump to the next level where we now see level 2 and 3.

This way, we no longer need to keep in sync the main navigation and the "sub" navigation. but instead have one navigation tree that we can "go deeper" into.

src/sidebar.ts
```diff
@@ -97,149 +106,151 @@ const quickstartSidebar = {
       collapsible: false,
       link: {
         type: "doc",
-        id: "getting-started/ory-network-oauth2",
+        id: "guides/oauth2-openid-connect",
```
aeneasr, line 109: Changed this because `getting-started/ory-network-oauth2` has a custom sidebar label, so now we added an autogenerated page

src/sidebar.ts
```diff
@@ -369,519 +392,524 @@ const guidesSidebar = {
...
+    collapsible: false,
+    link: {
+      type: "doc",
+      id: "kratos/emails-sms/custom-email-templates",
+    },
+    items: ["kratos/emails-sms/custom-email-templates"],
+  },
+]
+
+const oauth2Sidebar = [
```
aeneasr, line 514: I was not able due to time constraints to do the flattening option for all nav items, maybe you could follow up on this @vinckr ?

src/sidebar.ts
```diff
@@ -989,13 +1001,15 @@ module.exports = {
...
   ],
-  ...referenceSidebar,
-  ...quickstartSidebar,
-  ...guidesSidebar,
-  ...identitiesSidebar,
-  ...oauth2Sidebar,
-  ...permissionsSidebar,
-  ...selfhostingSidebar,
+  reference: referenceSidebar,
+  "quickstarts-and-tutorials": quickstartSidebar(false),
```
aeneasr, line 1005: And here the unflattened - so all levels - sidebar which is visible after we click on a category.

aeneasr:
This adds more content to the main navigation, and I think we can still merge a couple of categories together (e.g. identity schema and identity management).

docs/guides/authentication.mdx
vinckr: lgtm

src/sidebar.ts
vinckr: oh very cool, did not know this feature :D 

so basically the "flat" items we only show when you click on the category they are in, very nice

src/sidebar.ts
vinckr: so identitiesSidebar, oauth2Sidebar, permissionsSidebar, and selfhostingSidebar need flattening still? 

should I do it in this PR or do you want to merge asap?

src/sidebar.ts
```diff
@@ -97,149 +106,151 @@ const quickstartSidebar = {
...
-          type: "category",
-          label: "OpenID Connect SSO",
-          items: [
-            "kratos/social-signin/overview",
+      link: flat
+        ? {
+            type: "doc",
+            id: "guides/authentication",
+          }
+        : undefined,
```
vinckr, line 174:
```suggestion
      link:
         {
            type: "doc",
            id: "guides/authentication",
          },
```

Not quite sure what the undefined part does here? Did you remove the link on purpose? 

Can we link to the category overview doc here in the 2nd level? 

Like we do for Ory Actions for example.

Attached screenshot for clarification - I want every item in the category to be clickable/lead to a doc.
Screenshot 2024-09-23 at 11 50 44

src/sidebar.ts
aeneasr: Correct

Not really a future but a creative way to (ab)use the docusaurus nav bar :D

src/sidebar.ts
aeneasr: I think it's fine in another PR, this already improves the state I think

aeneasr: basically all sidebars that don't have a function call yet

src/sidebar.ts
aeneasr: Yes, I was experimenting how to get the links working and forgot to clean it up. thank you!. I cleaned it up!

PR #1893, opened 2024-09-24
chore(typo): Remove the extra 'is', and correct to Rules

oleorhagen:
Simple typo fix


vinckr (approved):
thanks for the fix! 

if you could sign the CLA it would satisfy our CI bot as well, thanks again 🙏

PR #1929, opened 2024-10-24
Update cache.mdx - filename typo

abooij:
The two surrounding snippets say `identity-config.yml`, but the file box itself is titled `config.yml`.


vinckr (approved):
thanks for spotting that!

PR #1936, opened 2024-11-05
docs: support for oid microsoft parameter

aeneasr:
@vinckr changelog pls

> In the case of Microsoft, using sub as an identifier can lead to problems. Because the use of OIDC at Microsoft is based on an app registration, the content of sub changes with every new app registration. Sub is therefore not uniquely related to the user. It is therefore not possible to transfer users from one app registration to another without further problems.
https://learn.microsoft.com/en-us/entra/identity-platform/id-token-claims-reference#payload-claims
> With the use of oid it is possible to identify a user by a unique id.

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [ ] I have referenced an issue containing the design document if my change introduces a new feature.
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security vulnerability. 
      If this pull request addresses a security vulnerability, 
      I confirm that I got approval (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added the necessary documentation within the code base (if appropriate).

PR #1952, opened 2024-11-19
feat: add docs for the use of oid within microsoft

teawithfruit:
Add docs for the use of OID within Microsoft.


teawithfruit:
Duplicate... #1936

PR #1985, opened 2024-12-25
feat: preload fonts and reduce icon size

aeneasr:
## Checklist


jonas-jonas (commented):
image
I don't think this works, unfortunately. Maybe it's just a thing with the preview, though.
image

aeneasr:
Strange, it was working just fine in dev mode. Maybe the plugin loading chain is messed up

aeneasr:
Pagespeed is on fleek now since we don't need to wait, download, parse, swap/render fonts :D

aeneasr:
Yeah, looks like in prod mode docusaurus does not care about my plugin, only in dev mode :(

aeneasr:
It's so messed up - you can't reference the fonts directly in CSS (it always bundles it with content hashes) and you can't change the bundling. So basically you can't pre-load fonts. Webfonts suck...

src/plugins/docusaurus-static-fonts/index.js
```diff
@@ -0,0 +1,25 @@
+const webpack = require("webpack")
+const NodePolyfillPlugin = require("node-polyfill-webpack-plugin")
+
+// eslint-disable-next-line
+module.exports = function (context, options) {
+  return {
+    name: "docusaurus-static-fonts",
+    // eslint-disable-next-line
+    configureWebpack(config, isServer, utils) {
```
jonas-jonas, line 9: Shouldn't these files just be statically served, as a static asset? Does webpack need to process them?

aeneasr: No it doesn't, but I can't figure out how to tell docusaurus to NOT modify them. As soon as CSS uses `url` it bundles the files and ads content hashes. I will probably just reference the file with the content hash in preload as the contents of the file won't likely change anyways.

jonas-jonas: Ahh, webpack processes them. Okay I understand.. maybe we can just host these fonts on a static file server or using cloudflare on `ory.sh`

aeneasr: We could, but the CSS bundler does its thing and replaces the paths

PR #1988, opened 2024-12-25
revert: webfont loading

aeneasr:
## Checklist


jonas-jonas (commented):
You could also change the head tag to reference `https://www.ory.sh/fonts/Inter/InterVariable.woff2?v=4.0` directly.

aeneasr:
I tried basically everything but it was not possible to solve it because the CSS bundler automatically bundles (and builds and places and renames) the referenced font files.

What i have not tried is a static css file - maybe that works

jonas-jonas:
I found [this blog post](https://johnnyreilly.com/preload-fonts-with-docusaurus) and it mentioned this:

>Note that the urls are fully qualified to prevent webpack from trying to bundle them

So yea, seems to be the only option.

aeneasr:
Ah nice, thanks - FQDN did not work because of cors loading when running locally / in preview …

aeneasr:
But I think I can load an unbundled static font file which is not bundled, maybe that works

src/static/fonts/fonts.css
```diff
@@ -8,15 +8,15 @@
   font-style: normal;
   font-weight: 100 900;
   font-display: swap;
-  src: url("/fonts/Inter/InterVariable.woff2?v=4.0") format("woff2");
+  src: url("/docs/fonts/Inter/InterVariable.woff2?v=4.0") format("woff2");
```
aeneasr, line 11: Loading this statically allows us to define absolute paths and circumvent the webpack bundler

src/static/scripts/kapa.js
```diff
@@ -0,0 +1,17 @@
...
+script.setAttribute(
+  "data-modal-disclaimer",
+  "By utilizing this chatbot, you consent to the collection and transmission of data to kapa.ai, which may include your IP address. Please be advised that your privacy and data protection are of utmost importance to us. We assure you that any data collected will be handled in compliance with applicable laws and regulations. For further details on how your data is processed and used, we encourage you to review our Privacy Policy. If you do not agree with these terms, we kindly request that you refrain from using this chatbot.",
+)
+script.setAttribute("data-modal-title", "Ory AI Copilot")
+script.setAttribute("data-button-text", "Ask AI")
+script.setAttribute("data-project-logo", "/docs/img/kapa-logo.png")
+script.defer = true
+
+addEventListener("DOMContentLoaded", (event) => {
```
aeneasr, line 15: kapa is pretty expensive because they load a lot of things immediately. I've contacted them about it. But in the meanwhile, we only load it after the page is done and defer it a bit more to make it not impact page loading.

PR #1991, opened 2025-01-06
fix: handle auth0 export job state of "processing"

claym-ih:
For auth0 export jobs against connections with a large number of users, there's an additional job state, `processing`, which does not seem to be accounted for. 

This fix treats the handling of `processing` like `pending`, waiting for the job to finish before moving on to the next step.


aeneasr (approved):
Amazing, thank you!

aeneasr:
Thank you, this looks great! Could you please sign the CLA (just two clicks), then we can merge it right away!

PR #2031, opened 2025-02-14
fix: move community projects to github

vinckr:
moving community content from docs to https://github.com/ory/awesome-ory


docs/ecosystem/community.md
```diff
@@ -9,186 +9,17 @@ Sponsors, contributors, and adopters make this ecosystem possible and shape it i
 Take part in the Ory Community!
 
 - Chat on the [Ory Community Slack](https://slack.ory.sh/)
-- Discuss or share code on [GitHub](https://github.com/orgs/ory/discussions)
 
-Each major project also has its own discussion board:
```
aeneasr, line 14: Should we keep these links here for people to engage? Otherwise this page is pretty empty and can probably removed all together

aeneasr: It moved to the bottom, I see

vinckr: yea we can think about if we want to keep this page - or rather merge it with one of the other ones 🤔 have not had a good idea yet. 

right now it would mostly link to the repo containing all the links to community projects

aeneasr (approved):

PR #2049, opened 2025-03-06
docs: update node express quickstart

christiannwamba:
## Checklist


code-examples/protect-page-login/expressjs/index.js
```diff
@@ -0,0 +1,22 @@
+const app = require("express")()
+// highlight-next-line
+const sdk = require("@ory/client-fetch")
+
+// highlight-start
+const ory = new sdk.FrontendApi(
+  new sdk.Configuration({
+    basePath: process.env.ORY_ENDPOINT || "http://localhost:4000/.ory",
```
aeneasr, line 8: We use `ORY_SDK_URL` consistently across the docs, can you use that here too please? :)

code-examples/protect-page-login/expressjs/index.js
```diff
@@ -0,0 +1,22 @@
...
+  }),
+)
+ // highlight-end
+
+app.get("/", (req, res) =>
+  // highlight-start
+  ory
+    .toSession({ cookie: req.header("cookie") })
+    .then((data) => res.json(data))
```
aeneasr, line 17: I like the minimalism of the example, but maybe something that is a tiny bit more human-readable would be nice here? :)

code-examples/protect-page-login/expressjs/index.js
```diff
@@ -0,0 +1,22 @@
...
+    .catch(() => res.redirect("/.ory/ui/login")),
+  // highlight-end
```
aeneasr, line 19: This can be as is, but realistically users will use Ory like a middleware so like:

```
app.get("/", requireAuth((req, res) => {
  // req.session.id
})
```

For the sake of getting this over the line and since it was like this before I think it's fine as is, but it probably isn't what people would need in a real world application.

christiannwamba: Got it. I think we should use middlewares since its the express standard. Will update.

code-examples/protect-page-login/expressjs/index.js
christiannwamba: Makes sense. Will update :)

piotrmsc (approved):

jonas-jonas (approved):

PR #2073, opened 2025-03-19
chore: redirects for 404 or moved pages

vinckr:
fixing some redirects for 404 pages.


christiannwamba (approved):
Thanks for catching these @vinckr. Looks great!

PR #2087, opened 2025-03-28
DOCUMENTATION: Pin go sdk version to 2.2.1 since there is not a 2.3.0 version released.

sethtomy:
DOCUMENTATION: Pin go sdk version to 2.2.1 since there is not a 2.3.0 version released.

## Checklist

- [n/a] I have referenced an issue containing the design document if my change introduces a new feature.
- [n/a] I have added tests that prove my fix is effective or that my feature works.
- [n/a] I have added the necessary documentation within the code base (if appropriate).

## Further comments

I ran into this when using the documentation to use the go sdk.

If it's preferred to regenerate and publish the client (I'm sure it is), a small pointer in the right direction I'm happy to do that instead.

### DIFF
--- a/docs/hydra/self-hosted/05_go.mdx
+++ b/docs/hydra/self-hosted/05_go.mdx
@@ -34,7 +34,7 @@ To install the Go SDK, run:
 import CodeBlock from '@theme/CodeBlock'
 import { useLatestRelease } from '@site/src/hooks'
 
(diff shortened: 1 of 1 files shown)


vinckr:
Hey @sethtomy 

thanks a lot! 

I agree that getting the new version out would be the preferred option... you can maybe find some pointers here: https://github.com/ory/sdk - personally I also have never done it and most of the dev staff is already in the weekend. I will see to get some eyes on this after the weekend., 
thanks

PR #2126, opened 2025-04-22
docs: account enumeration mitigation

aeneasr:


docs/identities/sign-in/identifier-first-authentication.mdx
```diff
@@ -7,6 +7,12 @@ sidebar_label: Identifier first authentication
 Identifier first authentication first requests the user's identifier such as an email or username before prompting for a password
 or other authentication methods.
 
+:::note
+
+Identifier first authentication is required when using B2B Organization login.
+
+:::
```
jonas-jonas, line 14: This note should be on the orgs docs, I think.

docs/identities/sign-in/identifier-first-authentication.mdx
```diff
@@ -35,3 +41,23 @@ selfservice:
...
 
 To disable this feature, set `style` to `unified`.
+
+## Account enumeration mitigation
+
+Account enumeration mitigation prevents malicious actors from being able to identify if a user exists or not.
+
+Per default, account enumeration mitigation is disabled when using identifier first authentication. This improves user experience
+as the user quickly sees if they are signed up or not with the chosen identifier (email / username). To enable account
+enumeration, use the Ory CLI patch command
```
jonas-jonas, line 51:
```suggestion
By default, Ory does not prevent account enumeration in the identifier first authentication flow. This improves user experience
as the user quickly knows if they have an account with the chosen identifier (email / username) or not. To enable account
enumeration, use the Ory CLI patch command
```

docs/identities/sign-in/identifier-first-authentication.mdx
aeneasr: I think in both, no?

aeneasr: Well, ideally this would be in the product itself and does not require the user to know what knobs to turn!

aeneasr: added the note

PR #2181, opened 2025-05-27
docs: onboarding portal

niwsa:


vinckr (approved):
LGTM!

Only nit is the (preview) in the heading [here](https://github.com/ory/docs/pull/2181/files#diff-e22fa24946003843673ca5d8d66c966084fba6bab448424fcba0713574c84cb7R438) - I would omit that since we often forget to remove it... 
better to have the tag in the console (as the SS shows) and then also mention it in the note there. But its a nit so not super important...

I added the `upstream` label since I take this is waiting for a release still

docs/kratos/organizations/organizations.mdx
```diff
@@ -389,3 +389,157 @@ local claims = std.extVar('claims');
...
+### Streamline organization onboarding
+
+The onboarding portal addresses the often-burdensome process of setting up and managing identity connections with external
+organizations. Traditionally, integrating with a customer's or partner's identity provider (IdP) for single sign-on (SSO) or
+synchronizing user directories involves manual, error-prone configurations. This often requires deep technical expertise in
+protocols such as SAML.
+
+Onboarding portal aims to:
+
+- **Simplify complex integrations**: By providing a self-service or guided onboarding experience, they abstract away the
```
vinckr:
```suggestion
- Simplify complex integrations**: By providing a self-service or guided onboarding experience, they abstract away the
```

please use bolding __only__ for UI elements

niwsa: You are right. This is not connected to the polis admin portal. The onboarding portal is meant for our customer's customer to set up SSO without access to the ory network console.

docs/kratos/organizations/organizations.mdx
```diff
@@ -389,3 +389,157 @@ local claims = std.extVar('claims');
...
+  (for example, Microsoft Entra ID, Okta, Google Workspace) for seamless access to the SaaS application.
+- Enterprise IT administrators: These individuals in your customer organizations are responsible for managing their internal
+  identity providers and user directories. They use the onboarding portal to configure the connection to your application.
+- Developers and architects building multi-tenant applications: These individuals implement robust and scalable identity
+  management for applications that serve multiple organizations. Each organization has its own authentication and user
+  provisioning requirements.
+
+Current limitations:
+
+- The initial version includes a guided walkthrough for a limited set of providers. OIDC support and additional providers are
```
piotrmsc: I would specify that we currently support SAML providers and list the ones if we know are not compliant to be more transparent.

docs/kratos/organizations/organizations.mdx
```diff
@@ -389,3 +389,157 @@ local claims = std.extVar('claims');
...
+</BrowserWindow>
+```
+
+```mdx-code-block
+</TabItem>
+<TabItem value="api" label="API">
+```
+
+#### Create an onboarding portal link
+
```
piotrmsc, line 482: Does the authentication work with Project API Key? If yes it should be described and in both support then update the matrix : https://www.ory.sh/docs/concepts/personal-access-token#workspace-api-keys-vs-project-api-keys this

Plus as prereq link to https://www.ory.sh/docs/concepts/personal-access-token because user needs to have API Key in general.

docs/kratos/organizations/organizations.mdx
```diff
@@ -389,3 +389,157 @@ local claims = std.extVar('claims');
...
+    }
+  ]
+}
+```
+
+```mdx-code-block
+</TabItem>
+</Tabs>
+```
+
```
piotrmsc, line 547: I would also mention later the handover of the link to respective audience and I would show how it looks like - the actual screen and mention what happens when link is re-used /re-opened.

piotrmsc (approved):

PR #2255, opened 2025-07-25
feat: add inflight rate limiting documentation and improve tooling

tricky42:
- Add comprehensive endpoint-based rate limiting documentation
- Document inflight request limiting for concurrent request protection
- Add protected endpoints table with rate limit types and actions
- Add Makefile help target and dev server target
- Update dependencies to latest minor versions


docs/guides/rate-limits.mdx
```diff
@@ -162,58 +162,89 @@ Endpoint based rate limits are controls applied to individual API endpoints with
...
    - Contributes to a fairer and more stable platform for all users by mitigating abusive traffic.
 
-5. Flexibility for high volume legitimate traffic:
+5. **Flexibility for high-volume legitimate traffic:**
    - Provides options for Enterprise and Growth customers to whitelist internal IPs for higher rate limits.
    - Balances security needs with the requirements of high-volume legitimate traffic.
 
-### Notes on rate limit rules
+### Types of endpoint-based protection
+
```
Demonsthere, line 175: I think here we should keep the distinction of layers:
1. Global CDN layer with Cloudflare
2. Local Business logic layer with Envoy
The current proposal flattens that which i don't think is a good idea.

tricky42: I intentionally chose not to disclose these "implementation" details, as they shouldn't affect the customer, and it also allows us the flexibility to modify the implementation without needing to update this documentation or notify the customers.

Demonsthere (approved):

PR #2261, opened 2025-07-30
docs: expand docs for SMS recovery

gaultier:


docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
 
 ### Enable SMS account recovery
 
+This is enabled by turning a feature flag to 'on' at the project level (opt-in). This flag is enabled by default for newly created
```
vinckr:
```suggestion
This feature is enabled via a project-level feature flag, which is `on` by default for new projects in Ory Network.
```

can you please also provide the link to where you enable this in the console for existing projects?

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
...
+projects.
+
+This feature only works when the recovery strategy is `code` (not `link`).
+
+#### As a user
+
+If you only have an email address and no phone number, there is no perceivable difference.
```
vinckr:
```suggestion
For users with only an email address and no phone number as identifier, the recovery process is unchanged.
```

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
...
+
+If you do have a phone number and want to receive a SMS containing the recovery code, make sure that this phone number is
+registered as a recovery address.
```
vinckr:
```suggestion
To receive a recovery code via SMS, the users phone number must be registered as a recovery address.
```

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
...
+
+If you have multiple recovery addresses registered, during the recovery flow, you will be presented a list of your (masked)
+addresses and you will be prompted to choose one of them.
```
vinckr:
```suggestion
If the user has multiple recovery addresses registered, they will be prompted to choose one from a masked list during the recovery process.
```

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
...
+
+You may also be prompted to enter this address in full to prove that you know the complete address, for security reasons.
```
vinckr: what means "may" here? is the user prompted to provide the full address or not @gaultier

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,25 @@ configure account recovery in your Ory project.
...
+
+#### As an Administrator
```
vinckr, line 64: what is the difference between developer and administrator here @gaultier ?

vinckr: just for me to understand what is the difference between developer and administrator here / why do they have a separate section @gaultier. 
i don't think I saw this distinction in other places in the docs before.

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
 4. Toggle "Receive a recovery code via SMS" to 'on'
 5. Click **Save**
 
+It is safe to toggle this feature flag on and off. This flag only is take into account when a new recovery flow is created.
+Existing flows will be unaffected by toggling the flag.
```
vinckr:
```suggestion
It is safe to toggle this feature flag. Changes only take effect when a new recovery flow is initiated; existing flows are unaffected.
```

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
+
+SMSes can only be delivered if a [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) is configured.
```
vinckr:
```suggestion
A [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) must be configured to deliver SMS messages.
```

docs/identities/get-started/account-recovery.mdx
vinckr:
```suggestion
#### As an administrator
```

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
+
+#### As a developer
+
+When using `elements`, there is nothing special to do. Otherwise, you need to do some changes.
+
+New fields have been added to the payload when getting and updating the recovery flow that are exclusively used when this feature
+flag is enabled:
```
vinckr, line 98:
```suggestion
When this feature is enabled, the following fields are added to the recovery flow payload:
```

is that correct? in the previous version it's not clear to me if the fields are added always (but not always used) or just when the flag is enabled

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
 5. Click **Save**
 
+It is safe to toggle this feature flag on and off. This flag only is take into account when a new recovery flow is created.
+Existing flows will be unaffected by toggling the flag.
+
+SMSes can only be delivered if a [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) is configured.
+
+#### As a developer
+
+When using `elements`, there is nothing special to do. Otherwise, you need to do some changes.
```
vinckr: not sure if I understand this sentence. 
You mean if you build a custom UI using Ory Elements there is nothing special to do? 
What about the built-in Account Experience UI - I assume you also don't need to do anything and you only need to change things if you use a direct API integration for custom UI? 
did I get that right @gaultier ?

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
+New fields have been added to the payload when getting and updating the recovery flow that are exclusively used when this feature
+flag is enabled:
+
+- `recovery_address`
+- `recovery_select_address`
+- `recovery_confirm_address`
+- `screen`
+
+When integrating with the API directly, your applications must send and receive these fields. Refer to the OpenAPI specification
+for more details.
```
vinckr:
```suggestion
When integrating directly with the API, your application must send and receive these fields.
```

Please add a direct link to the API reference here as well.

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +71,28 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
+
+There is a limit of 10 recovery addresses to be considered and shown during the recovery flow for a particular user.
```
vinckr:
```suggestion
A maximum of 10 recovery addresses will be shown to a user during the recovery flow.
```

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -160,12 +160,23 @@ In the Ory Console go to <ConsoleLink route="project.mfa" />
 ### Account recovery address
 
 To start account recovery, Ory Identities must know which address to send the recovery message to. Usually this is the email
-address the user provides when registering their account. Other fields inside the `traits` section are supported as well.
+address the user provides when registering their account. Other fields inside the `traits` section are supported as well such a
+phone number to receive the code via a SMS.
```
vinckr:
```suggestion
address the user provides when registering their account. Other fields inside the `traits` section are supported as well such a
phone number to receive the code via a SMS.
```

what other fields besides phone number are supported? 
if only email and phone number are supported please state it clearly.

docs/identities/get-started/account-recovery.mdx
gaultier: > can you please also provide the link to where you enable this in the console for existing projects?

this is present a few line below

docs/identities/get-started/account-recovery.mdx
gaultier: Not always. The flow goes like this:

1. Ask for any recovery address. If this address does not exist at all or if it exists and it's the only one for the user, go to 4.
2. If there are multiple address, present a choice.
3. **If the choosen address on screen 2 is different from the one provided on screen 1, ask for the full address.**
4. Ask for the code that was sent to the address

So only if there are multiple addresses *and* the one entered on screen 1 is different from the one picked on screen 2, will the user be asked to enter the address fully. 

See the state `recovery_confirming_address` on the (admittedly complex) [diagram](https://github.com/ory-corp/cloud/blob/9d04ef63b6f44cd37772f2f5728c408e0500dbdc/kratos/kratos-oss/selfservice/flow/state_recovery_v2.mermaid#L1-L1).

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -160,12 +160,23 @@ In the Ory Console go to <ConsoleLink route="project.mfa" />
...
+
+If there are multiple recovery addresses registered for the user and if the feature flag `choose_recovery_address` is
+[enabled](../../../identities/get-started/account-recovery.mdx#enable-the-feature-flag-choose_recovery_address-unlocks-sending-a-recovery-code-via-sms),
+entering any of them will present them a list of (masked) addresses to choose from (up to 10). Once they have picked one address,
+and entered the address fully (to demonstrate that they indeed know the address and also to avoid information exfiltration), this
+address is used. The address can be an email address, in which case an email is sent, or a phone number, in which case a SMS is
+sent.
```
vinckr:
```suggestion
```suggestion
If a user has multiple recovery addresses and the `choose_recovery_address` feature flag is [enabled](../../../identities/get-started/account-recovery.mdx#enable-the-feature-flag-choose_recovery_address-unlocks-sending-a-recovery-code-via-sms), the process is as follows:

1. The user enters any of their registered addresses to begin the recovery flow.
1. A masked list of their recovery addresses is displayed. Up to 10 addresses are shown.
1. After selecting an address, the user must enter it in full. This step verifies ownership and helps prevent information leaks.
1. The recovery message is sent to the confirmed address via email or SMS.
```
```

File: docs/kratos/self-service/flows/account-recovery-password-reset.mdx [hunk 15]
```diff
@@ -160,12 +160,23 @@ In the Ory Console go to <ConsoleLink route="project.mfa" />
...
+
+If you support phone numbers in your identity schema, be sure to configure a
+[SMS HTTP gateway](../../emails-sms/10_sending-sms.mdx) so that SMSes are indeed sent.
```
vinckr:
```suggestion
	A [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) must be configured to deliver SMS messages.
```

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -160,12 +160,23 @@ In the Ory Console go to <ConsoleLink route="project.mfa" />
...
 
 :::info
 
-If the email address used for recovery is the same as the email used for verification and the account isn't activated when the
-recovery flow is started, completing the recovery flow also verifies the user's email address.
+If the address used for recovery is the same as the address used for verification and the account isn't activated when the
+recovery flow is started, completing the recovery flow also verifies the user's address.
```
vinckr, line 180:
```suggestion
Completing a recovery flow will also verify the identities address if the address is used for both recovery and verification and the identity is not yet verified when the recovery flow begins.
```

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -204,14 +215,79 @@ To specify a trait of the identity to be used for recovery, use the following id
...
 When this option is on and users attempt to initiate recovery for unregistered addresses, the system sends an attempted recovery
 notification to the email address that was used in the attempt. This prevents account enumeration attacks as explained in this
 [blog post by Troy Hunt](https://www.troyhunt.com/website-enumeration-insanity-how-our-personal-data-is-leaked/). By default, this
 feature is disabled in newly created Ory Network projects.
 
-Follow these steps to enable sending attempted recovery notifications:
+Note that for costs reason, this concept does not exist for SMSes. That means that trying to recover an account with the wrong
+phone number will _not_ result in any sent SMS.
+
+Follow these steps to enable sending attempted recovery notification emails:
```
vinckr:
```suggestion
For cost reasons, the attempted recovery notification does not apply to SMS. An attempt to recover an account using an incorrect phone number will not trigger an SMS message.

Follow these steps to enable sending attempted recovery notification emails:
```

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -324,11 +401,10 @@ Learn how to [add and configure hooks for self-service user flows](../../hooks/0
...
 
-In some scenarios, users may want to recover their account using a different email address than the one they used to register, for
-example if they no longer have access to the original email address. To use two (or more) recovery email addresses, you need to
-define a secondary email field, such as `email_secondary` in the traits section of the identity schema. This field will serve as
-the fallback email address for account recovery. Read more about identity schemas
-[here](../../manage-identities/05_identity-schema.mdx).
+In some scenarios, users may want to recover their account using a different address than the one they used to register, for
+example if they no longer have access to the original address. To use two (or more) recovery addresses, you need to define a
+secondary field, such as `email_secondary` in the traits section of the identity schema. This field will serve as the fallback
+address for account recovery. Read more about identity schemas [here](../../manage-identities/05_identity-schema.mdx).
```
vinckr:
```suggestion
A user may need a fallback recovery address if they lose access to the address used for registration.

To enable this, define a secondary field, such as email_secondary, in the traits section of your [identity schema](../../manage-identities/05_identity-schema.mdx). This field will serve as the alternative address for account recovery.
```

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -400,6 +476,10 @@ In this schema, you define both `email` and `email_secondary` fields as recovery
 messages to both addresses when the recovery flow is initiated. The user should provide the `email_secondary` field during
 registration or update it later in their account settings so it can be used in the recovery process.
 
+Note that this feature is orthogonal to the situation where a user has multiple recovery addresses registered of different
+_types_, such as one email address and one phone number. In this case they will be prompted to choose which address to use during
+the recovery flow and only _one_ message will be sent to the address they picked.
+
```
vinckr:
```suggestion
This is different from the scenario where a user has multiple recovery addresses of different types, such as one email address and one phone number. In that case, the user is prompted to choose which address to use, and a recovery message is sent to only the one they select.
```

vinckr (changes requested):
Thanks looking good already. 

I revised the text a bit to make it more readable and left some comments where I was not sure. 

@unatasha8 feel free to also take a final look after my comments have been resolved. 

thanks!

vinckr:
sorry for some reason I submitted my partial review and then the full one. 
Ill mark the comments as resolved that you already addressed.

docs/identities/get-started/account-recovery.mdx
vinckr: So basically if the user inputs their address in step 1 and then chooses that address in step 3 they wont be asked to input the full address bc they did already, that makes sense. 

how about 
```suggestion
To confirm ownership, the user may be asked to enter the complete address after selecting it. This verification step is skipped if they already entered that same address to begin the recovery process.
```

docs/identities/get-started/account-recovery.mdx
gaultier: Right, in my mind an administrator is someone who can change the settings in the console (or the configuration), toggle feature flags, etc. Does not have to be technical. More like an IT or backoffice person.

A developer is the one integrating our SDK and API against their own APIs and apps.

Not sure if it makes sense to have that distinction?

vinckr: Yea I think we can have it in one section for now as most Ory Network users are both admin and dev in the same person 😁

docs/identities/get-started/account-recovery.mdx
gaultier: Your suggestion is correct. Also, when the feature flag is enabled, the old fields are ignored (it really is a different payload in the end, with no overlap with the previous one.)

docs/identities/get-started/account-recovery.mdx
gaultier: I merge them then

docs/identities/get-started/account-recovery.mdx
gaultier: I mentioned the ignored fields

docs/identities/get-started/account-recovery.mdx
gaultier: I think I meant here Account Experience, not elements. Will fix.

docs/identities/get-started/account-recovery.mdx
gaultier: Done

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
gaultier: The identity schema can have arbitrary fields but the recovery can currently only be done via email or sms, and the mapping from identity trait to recovery channel  (email/sms) is done in the identity schema.

vinckr (approved):
awesome, thanks for your patience and addressing the many comment so quickly 😁 

its looks good from my POV, lets have @unatasha8 take a look before we merge though

unatasha8:
Will take a look after a couple of meetings I have.

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
 
 ### Enable SMS account recovery
 
+This feature is enabled via a project-level feature flag, which is `on` by default for new projects in Ory Network. projects.
```
unatasha8: Looks like an errant period "Ory Network. projects."

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
...
+
+This feature only works when the recovery strategy is `code` (not `link`).
```
unatasha8: I'm presuming that 'code' refers to the one-time password option as that is the only other option to magic link. In which case: 
"SMS account recovery is only available when you enable the one-time passwords recovery option."

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
...
+
+#### As a user
+
+For users with only an email address and no phone number as identifier, the recovery process is unchanged.
```
unatasha8: "For users that use only email as their identifier (a phone number was not provided), the recovery process remains the same."

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
...
+
+To receive a recovery code via SMS, the users phone number must be registered as a recovery address.
```
unatasha8: Add apostrophe "the user's phone number"

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
...
+
+If the user has multiple recovery addresses registered, they will be prompted to choose one from a masked list during the recovery
+process.
+
+To confirm ownership, the user may be asked to enter the complete address after selecting it. This verification step is skipped if
+they already entered that same address to begin the recovery process.
+
+#### As a developer
+
 Enable the feature flag `choose_recovery_address` to be able to send a recovery code via SMS.
```
unatasha8: switch order of 'feature flag' and `choose_recovery_address` 
remove 'be able to'
remove 'via'
"Enable the `choose_recovery_address` feature flag to allow sending a recovery code using SMS."
Question: Why refer to the feature flag (I presume this is the flag name in code) if the instructions are to use the console UI? Do they need to know it for using the CLI? Are we co-mingling the Console and UI instructions here?

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +70,32 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
 4. Toggle "Receive a recovery code via SMS" to 'on'
 5. Click **Save**
 
+It is safe to toggle this feature flag. Changes only take effect when a new recovery flow is initiated; existing flows are
+unaffected.
+
+A [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) must be configured to deliver SMS messages.
```
unatasha8: It seems a bit late to provide this requirement. Perhaps move it to a more up front location in the doc?

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +70,32 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
 4. Toggle "Receive a recovery code via SMS" to 'on'
 5. Click **Save**
 
+It is safe to toggle this feature flag. Changes only take effect when a new recovery flow is initiated; existing flows are
```
unatasha8: Do you mean when a new recovery flow is added (by a developer) to the system or when a user initiates a new recovery flow?

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +70,32 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
+- `screen`
+
+And these fields will have no effect anymore (they can be present in the payload, but will be ignored):
+
+- `email`
+
+When integrating with the API directly, your applications must send and receive these fields. Refer to the
+[API documentation](../../reference/api#tag/frontend/operation/updateRecoveryFlow) for more details.
+
+A maximum of 10 recovery addresses will be shown to a user during the recovery flow.
```
unatasha8: If this is possible instead of certain, change 'will' to 'can'
"A maximum of 10 recovery addresses can be shown to a user during the recovery flow."

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -160,12 +160,24 @@ In the Ory Console go to <ConsoleLink route="project.mfa" />
...
+1. After selecting an address, the user must enter it in full. This step verifies ownership and helps prevent information leaks.
+1. The recovery message is sent to the confirmed address via email or SMS.
+
+A [SMS HTTP gateway](../../emails-sms/10_sending-sms.mdx) must be configured to deliver SMS messages.
 
 :::info
 
-If the email address used for recovery is the same as the email used for verification and the account isn't activated when the
-recovery flow is started, completing the recovery flow also verifies the user's email address.
+Completing a recovery flow will also verify the identities address if the address is used for both recovery and verification and
```
unatasha8: I prefer the orginal sentence. Just replace 'email address' with 'address'
"If the address used for recovery is the same as the address used for verification and the account isn't activated when the recovery flow is started, completing the recovery flow also verifies the user's address."

docs/kratos/self-service/flows/account-recovery-password-reset.mdx
```diff
@@ -204,14 +216,79 @@ To specify a trait of the identity to be used for recovery, use the following id
...
 When this option is on and users attempt to initiate recovery for unregistered addresses, the system sends an attempted recovery
 notification to the email address that was used in the attempt. This prevents account enumeration attacks as explained in this
 [blog post by Troy Hunt](https://www.troyhunt.com/website-enumeration-insanity-how-our-personal-data-is-leaked/). By default, this
 feature is disabled in newly created Ory Network projects.
 
-Follow these steps to enable sending attempted recovery notifications:
+For cost reasons, the attempted recovery notification does not apply to SMS. An attempt to recover an account using an incorrect
+phone number will not trigger an SMS message.
+
+Follow these steps to enable sending attempted recovery notification emails:
```
unatasha8: The 'attempted recovery notification' noun phrase is so long... simplify the adverb to:
"Follow these steps to send attempted recovery notification emails:"

docs/identities/get-started/account-recovery.mdx
```diff
@@ -52,6 +70,32 @@ Enable the feature flag `choose_recovery_address` to be able to send a recovery
...
 5. Click **Save**
 
+It is safe to toggle this feature flag. Changes only take effect when a new recovery flow is initiated; existing flows are
+unaffected.
+
+A [SMS HTTP gateway](../../kratos/emails-sms/10_sending-sms.mdx) must be configured to deliver SMS messages.
+
+##### API changes
+
+When using the built-in Account Experience UI, there is nothing special to do. Otherwise, you need to do some changes.
```
unatasha8, line 95: Do you mean that if the developer/admin changes the recovery method in the console UI, then they do not need to make any API changes? 
Is this section an alternate (API) method to change the recovery method?

unatasha8 (commented):
While this PR doesn't add or alter the 'one-time codes' and 'magic links' definitions, we should clean up their usage in the 2 articles. The console UI calls it 'One-time passwords' but this article uses 'code' to refer to it. 

For example: 
### User flow
When the feature flag `choose_recovery_address` is enabled and the `code` method is in use.

Is 'code' really the name of a API/method? It has the code styling.

docs/identities/get-started/account-recovery.mdx
```diff
@@ -44,6 +44,24 @@ configure account recovery in your Ory project.
 
 ### Enable SMS account recovery
```
unatasha8: This title might mislead readers. This enables a user choice. They can choose to recover using email or SMS, instead of only being able to use email. It's not enabling only SMS, which could be implied by this title. Options:
"Enable users to choose recovery method"
"Enable email and SMS account recovery"

docs/identities/get-started/account-recovery.mdx
gaultier: > Question: Why refer to the feature flag (I presume this is the flag name in code) if the instructions are to use the console UI? Do they need to know it for using the CLI? Are we co-mingling the Console and UI instructions here?

There is a section in the Console UI to toggle feature flags. That applies to 'network' customers (using the SaaS).
An alternative way for these customers to toggle that is with an API call (could be from their own software or with the `ory` CLI, which does this API call).

But some customers also run the EOL or OSS version, where the feature flag is defined in a configuration file for the Identities (Kratos) service (a YAML file).

Does this help?

docs/identities/get-started/account-recovery.mdx
gaultier: The latter, the former is not a thing in our system. Adding a new flow can only be done by us by writing code.

docs/identities/get-started/account-recovery.mdx
gaultier: Hmm not sure... What I want to express is: There is a hard-coded limit in the system so that up to 10 recovery addresses are retrieved from the database and returned to the browser.

docs/identities/get-started/account-recovery.mdx
gaultier: No, I meant: when enabling the feature flag,  additional fields are expected for the same API that does the recovery.

'Account experience' is this library/application that we provide (but customers do not have to use it), which has already been adapted to send/receive the new fields. It acts as a intermediary between the customer and our API.

But if a customer wants to use the API directly, then they are responsible for sending/receiving these new fields.

gaultier:
> Is 'code' really the name of a API/method? It has the code styling.

In a sense yes, it is the recovery strategy, and can be `code` or `link`. Maybe it does not have to be styled that way, we could just write: `when the one-time password recovery strategy is in use, ...`

docs/identities/get-started/account-recovery.mdx
gaultier: 100% correct, this feature flag actually does both. What about: "Enable users to choose recovery method and receive a recovery code SMS" ?

deepakprabhakara:
@gaultier Just realised we should document the config for Kratos as well, unless I missed it?

docs/identities/get-started/account-recovery.mdx
unatasha8: I think your suggested title is too long. However, if we want to include SMS in the title we also have to include email. How about:
"Enable users to choose email or SMS recovery method" I think it's obvious they will receive an SMS message with the code.

docs/identities/get-started/account-recovery.mdx
unatasha8: OK, I think are saying the same thing. There are 2 ways to turn on/implement this feature; 1 is via the Console/Account Experience, and 2 is via code and API changes.

For option 1, everything is implemented and there is nothing to do except turning it on.
For option 2, the developer must "Enable the feature flag `choose_recovery_address` ..." AND add new fields to the payload.

If the above is true then... 
A. The title 'As a developer' should be 'As an administrator'. I'm presuming here that a developer would not be using the Console, rather some security admin person. (This also seems to imply they need to be using Ory Network.)
B. 'Enable the `choose_recovery_address` feature flag to send a SMS containing a recovery code.' might not make much sense to the admin person. Perhaps something like... 'Enable users to choose the SMS recovery method and receive an SMS containing a recovery code.'
C. The title 'API changes' could now be "As a developer".
D. The first sentence should not mention the UI. Instead, use "Enable the `choose_recovery_address` feature flag to send a SMS containing a recovery code." We don't include how that can be done via code, so should add that instruction.
E. I'd move "When integrating with the API directly, your applications must send and receive these fields. Refer to the [API documentation](../../reference/api#tag/frontend/operation/updateRecoveryFlow) for more details." higher up in the section as it is declaring the requirement to implement new fields.

There are a couple of ways to categorize these instructions... here we are using a person's role. But consider if using the type of implementation might be better. For example, if Ory Network, you can do it via the console, for OEL, you do it via the API, etc.

unatasha8 (commented):
Replied to 2 replies!

docs/identities/get-started/account-recovery.mdx
gaultier: Good points all around, let's split the explanations by type of product (EOL/OSS, Network). I think it makes more sense.

docs/identities/get-started/account-recovery.mdx
gaultier: Changed to: `are shown`

PR #2266, opened 2025-08-06
docs: mention that the new password must be different from the old password

gaultier:
Check implemented in https://github.com/ory-corp/cloud/pull/9076

### DIFF
--- a/docs/kratos/self-service/flows/user-settings.mdx
+++ b/docs/kratos/self-service/flows/user-settings.mdx
@@ -542,6 +542,10 @@ curl -s -X GET \
 }
 ```
 
+When changing the password, the new password must be different from the old password. Otherwise, an error of this kind will be shown: 'The new password must be different from the old password.'.
+
+This is to help users, for example in the event of a password leak, where they want to change their password, and unknowingly set it to the same value as the old password. In that case, this check will help them realize that they need to set their new password to a different value.
+
 ### Link and unlink from Google, Facebook, GitHub, ..., OpenID Connect / OAuth 2.0
 
 :::tip Before you start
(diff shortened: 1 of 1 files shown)


docs/kratos/self-service/flows/user-settings.mdx
```diff
@@ -542,6 +542,10 @@ curl -s -X GET \
 }
 ```
 
+When changing the password, the new password must be different from the old password. Otherwise, an error of this kind will be shown: 'The new password must be different from the old password.'.
+
+This is to help users, for example in the event of a password leak, where they want to change their password, and unknowingly set it to the same value as the old password. In that case, this check will help them realize that they need to set their new password to a different value.
```
vinckr:
```suggestion
The new password must be different from the current one. Attempting to reuse a old password will trigger an error: `The new password must be different from the old password.` 

This check ensures that a new, unique password is set, preventing the accidental reuse of a potentially compromised credential.
```

vinckr (changes requested):
nice!

rewrote it a bit, wdyt @unatasha8 ?

docs/kratos/self-service/flows/user-settings.mdx
```diff
@@ -542,9 +542,9 @@ curl -s -X GET \
 }
 ```
 
-When changing the password, the new password must be different from the old password. Otherwise, an error of this kind will be shown: 'The new password must be different from the old password.'.
+The new password must be different from the current one. Attempting to reuse a old password will trigger an error: `The new password must be different from the old password.` 
 
-This is to help users, for example in the event of a password leak, where they want to change their password, and unknowingly set it to the same value as the old password. In that case, this check will help them realize that they need to set their new password to a different value.
+This check ensures that a new, unique password is set, preventing the accidental reuse of a potentially compromised credential.
```
unatasha8: Shorten it to: "This ensures that a new, unique password is set, and prevents the reuse of a potentially compromised password."

vinckr (approved):
lgtm ☺️

PR #2269, opened 2025-08-08
chore: next.js dir clarification

Milk-Tea:
As a user following this quickstart, it was unclear that the guide uses an alias.
The config and middleware files were to be placed in the root, but if a user has chosen to use a src/ directory, then the user will run into problems.

This change is to prevent such problems from arising.


vinckr (approved):
thanks so much for your contribution 🚀 

I fixed the style a bit

PR #2302, opened 2025-09-23
chore: update max grace period

hperl:


docs/hydra/guides/graceful-token-refresh.mdx
```diff
@@ -49,7 +49,7 @@ In this command:
 :::info
 
 The grace period cannot exceed 5 minutes unless a count limit is also configured. If a count limit is configured, the grace period
-can be set to a maximum of 30 days.
+can be set to a maximum of 180 days.
```
aeneasr: Make this OEL only, note that it is super insecure, and only for explicit migration scenarios!

PR #2361, opened 2025-11-24
feat: add docs for ory elements v1.2

jonas-jonas:
## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive API reference documentation for `@ory/elements-react`, including type aliases, helper functions, and component signatures.
  * Updated component prop types for improved type safety and clarity.
  * Added documentation for new helper functions for UI node type checking and placeholder resolution.
  * Added documentation for the `useResendCode` hook.
  * Reorganized reference documentation structure for improved navigation.


docs/elements/guides/05_upgrading.mdx
```diff
@@ -1,9 +1,38 @@
 ---
-id: migration-0x
-title: Migration from Ory Elements 0.x to 1.x
-sidebar_label: Migration from 0.x to 1.x
+id: upgrade
+title: Upgrading Ory Elements
+sidebar_label: Upgrading
```
vinckr:
```suggestion
title: Upgrade Ory Elements
sidebar_label: Upgrade
```

docs/elements/guides/05_upgrading.mdx
```diff
@@ -1,9 +1,38 @@
...
 ---
 
+# Ory Elements 1.1.0 to 1.2.0 Upgrade Guide
```
vinckr:
```suggestion
# Ory Elements 1.1.0 to 1.2.0 upgrade guide
```

sentence casing please

docs/elements/guides/05_upgrading.mdx
```diff
@@ -1,9 +1,38 @@
...
+- Settings Pages:
+  - Each settings section now similarly receives the nodes with a `buttonProps` or `inputProps` prop, depending on the type of
+    node.
+
+## Fixes
+
+- If you were using the `RecoveryCodesSettings` component, there was a typo in the prop name for regenerating codes. The correct
+  prop name is now `regenerateButton` instead of `regnerateButton`.
+
+# Ory Elements 0.x to 1.x Upgrade Guide
```
vinckr: we already have a h1 above, should they be h2 instead?

vinckr (changes requested):
nice!
2 small style things

docs/elements/guides/05_upgrading.mdx
jonas-jonas: It's a different section of the doc, not a child.. 

https://stackoverflow.com/questions/503925/is-it-alright-to-use-multiple-h1-tags-on-the-same-page-but-style-them-different

vinckr: my counter would be our style guide: https://github.com/ory/docs?tab=readme-ov-file#headings

>The highest heading level in documents is h2. Use h1 only when you override the document title.

this would be the first document afaict where we have multiple h1. my main concern is uniformity so you don't have to think a lot about style when writing docs. 

we should bring back the markdown linter tbh to avoid these discussions maybe

jonas-jonas: The reason it's in the style guide is SEO, and the linked article says it's allowed, just don't overdo.

But, I'll change it...

docs/elements/guides/05_upgrading.mdx
jonas-jonas: Actually no, it reads very weird in the sidebar. Going to keep "Upgrading" there.

docs/elements/guides/05_upgrading.mdx
jonas-jonas: And actually, it doesn't make sense to change both to H2 either, because the document doesn't have any h1 then. 

Let's keep two h1. It just doesn't make any sense otherwise.

jonas-jonas: nvm nvm, the docusaurus styling capitalizes the first h1 on the page more than the others, so it looks like the second is h2.. this is horrible.. 

I'll do a general h1, and two h2.

vinckr: The current setup is really only for 1 h1, has been that ever since I can remember - I mean we can change it, but I'm reluctant to rely on advice from a 16 year old video...
if I search I see very different advice - for example here it says don't do it: https://www.boia.org/blog/multiple-h1-tags-are-bad-for-accessibility-and-seo

>The `<h1>` tag is intended to define the most important heading on a page — essentially, it says “this is what this page is about.” It’s often identical or nearly identical to the page title tag.

This always made sense to me.

Ultimately I don't really care, and it's probably fine either way, at least to Google SEO. 

My main point is that we should have "dumb", **easy to follow** rules for style. The more complex it gets, the easier it is to mess up, the more discussions are needed, and harder to codify into a linter as well.

vinckr (approved):
lgtm

docs/elements/guides/05_upgrading.mdx
vinckr: > And actually, it doesn't make sense to change both to H2 either, because the document doesn't have any h1 then.

it always has a h1. if you don't specify one the title automatically gets rendered as h1

PR #2375, opened 2025-12-04
feat: phrasing on OEL changelog when no changes are mentioned otherwise.

alnr:


src/components/ListChangelogs/ListChangelogs.jsx
```diff
@@ -115,7 +115,7 @@ export default function ListChangelogs({ dir }) {
             {isNoChanges ? (
               <>
                 <Heading as="h2">{version}</Heading>
-                <p>No changes requiring a changelog in the {version}.</p>
+                <p>This release contains bugfixes and minor improvements.</p>
```
aeneasr: Bugfixes should be documented, so let's go with:

```suggestion
                <p>This version contains minor changes and improvements, such as dependency updates, only.</p>
```

PR #2522, opened 2026-04-23
chore: add AGENTS.md

hperl:
Add AGENTS.md with guidance on where to put what.

## Summary by CodeRabbit

* **Documentation**
  * Added contributor-facing guidelines organizing docs by deployment model with rules for migrated content and a step-by-step workflow for updates.
  * Defined a pattern using shared reusable content plus per-deployment shell pages that appear in sidebars.
  * Established canonical URL/SEO conventions (default canonical with a self-hosted exception) and how to set canonical tags.
  * Added a pointer to the new guidelines from the contributor guide.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,194 @@
+# Updating Ory Documentation
+
+## Overview
+
+Ory documentation is being organized by **deployment model**:
+
+- **Ory Network** → `docs/network`
+- **Ory Enterprise License (OEL)** → `docs/oel`
+- **Ory Open Source (OSS)** → `docs/oss`
+
+## Current State (Transitional Phase)
+
+- Most content is still **co-mingled**
+- Only some sections (e.g. Quickstarts) use the new structure
+- The Docs team is actively splitting content
+
+### What you should do
+
+- **Non-migrated sections** → continue using the existing structure
+- **Migrated sections** → follow the new pattern
+- **Do not restructure content yourself** unless it has already been migrated
+
+## How to Tell if Content is Migrated
+
+A section is migrated if:
+
+- The **sidebar** shows routes like:
+  - `docs/network/...`
+  - `docs/oel/...`
+  - `docs/oss/...`
+
+OR
+
+- You **cannot find the page in the old structure** and it exists under one of
+  the deployment folders above
+
+## New Structure (Migrated Sections Only)
+
+### Shared Source Content
+
+- Location: `/src/components/shared/<product>/...`
+- Purpose:
+  - Reusable, deployment-agnostic content
+  - Written once, used across deployments
+
+### Deployment Shell Files
+
+Locations:
+
+- `docs/network`
+- `docs/oel`
+- `docs/oss`
+
+Shell files:
+
+- Import content from `/src/components/shared/...`
+- Represent a page for a specific deployment
+- Are the files added to sidebars
+
+---
+
+## Sidebars
+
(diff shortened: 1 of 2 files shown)


aeneasr (commented):
Can you add `CLAUDE.md` with contents `@AGENTS.md` so that claude picks up the agents file? It's the official workaround to get claude to read agents

vinckr (commented):
do we need to commit this here? 
wouldn't it be enough for those that actually work on the docs to have locally/internally?

hperl:
> do we need to commit this here? wouldn't it be enough for those that actually work on the docs to have locally/internally?

I strongly suggest putting this directly into the repo, so that everybody has the same, up-to-date rules all the time. The alternative is clearly worse: every dev needs to add this file to the repo locally, and keep it current with what we have in Notion.

Rather, we should embrace this approach and not just add information about the structure, but also tone of voice, style guides, etc. `CONTRIBUTING.md` has little content on that.

vinckr:
> > do we need to commit this here? wouldn't it be enough for those that actually work on the docs to have locally/internally?
> 
> I strongly suggest putting this directly into the repo, so that everybody has the same, up-to-date rules all the time. The alternative is clearly worse: every dev needs to add this file to the repo locally, and keep it current with what we have in Notion.
> 
> Rather, we should embrace this approach and not just add information about the structure, but also tone of voice, style guides, etc. `CONTRIBUTING.md` has little content on that.

the README.md has content on the style that should be followed. Couldn't you give that as context? 

The guidenace in the file committed here seems to be only for the ongoing migration and should not be relevant after that is done.  

in any case we need to have 1 source of truth for these rules and having multiple files (agents, readme) will make things more confusing not less. 

couldn't you point to the README like you did in Claude.md "see README.md" and then add these rules about the migration strategy there?

hperl:
> The guidenace in the file committed here seems to be only for the ongoing migration and should not be relevant after that is done.

There's important information in the doc that is not in the README, which made me want to add it in the first place. Specifically the whole section about the new structure is just documenting the structure, not the migration:

```
### Shared Source Content

- Location: `/src/components/shared/<product>/...`
- Purpose:
  - Reusable, deployment-agnostic content
  - Written once, used across deployments

### Deployment Shell Files

Locations:

- `docs/network`
- `docs/oel`
- `docs/oss`

Shell files:

- Import content from `/src/components/shared/...`
- Represent a page for a specific deployment
- Are the files added to sidebars

---

## Sidebars

- Each deployment has its **own sidebar**

When adding a page:

- Add the **shell file** (not the shared source file)
- Only include it in deployments where the feature exists

## Canonical URL Rules (SEO)

When the same content exists across multiple deployments, you must define a
canonical URL.

### Default (Most Cases)

**Ory Network is canonical**

- Network → self-canonical
- OEL → canonical points to Network
- OSS → canonical points to Network

### Self-Hosted Exception

If content is specific to **self-hosted (OEL/OSS)**: **OEL is canonical**

- OEL → self-canonical
- OSS → canonical points to OEL
- Network → usually not applicable
```


 
> in any case we need to have 1 source of truth for these rules and having multiple files (agents, readme) will make things more confusing not less.
> 
> couldn't you point to the README like you did in Claude.md "see README.md" and then add these rules about the migration strategy there?

We can also put everything in README, I'm fine with that. I just want all documentation about the documentation structure to live in the repo, to be consumed by humans and agents alike.

aeneasr (approved):

PR #2552, opened 2026-05-12
feat: new sdk reference page with snippets

wassimoo:
## Summary
Adds a new SDK reference page to the API docs, initially for Python, Go, and TypeScript.

Each operation now shows real, copy-pasteable SDK snippets (generated from the Ory SDKs) alongside cURL, instead of only auto-generated HTTP examples.

image

## How it works
- Spec is the source of truth: The SDK snippets live inside the OpenAPI spec (src/static/api.json) as `x-codeSamples` entries on each operation.
- Injection script: `scripts/inject-sdk-code-samples.mjs` reads the generated Ory SDKs from the sibling `ory/sdk` repo and injects the Python, Go, and TypeScript samples into api.json:

## Open questions
Right now the SDK samples are injected manually. We should automate this eg. the same way we currently update:
- A CI job in the SDK source repo runs the injection + regeneration whenever the SDKs/spec change.
- The bot opens a chore(docs): update of SDK reference PR against this repo with the regenerated api.json and reference docs.


zepatrik (commented):
Some of the colors don't work well with light mode.

Image

I was also able to crash the page when switching to python.

Image

Additionally, as a user I'd want to have a very good search on that page where I can:
1. free text say what I want to do, e.g. "create identity" or "import identity"
2. search by the SDKs method, e.g. "createIdentity"

We might not do the search right away, but IMO we should have it soon.

unatasha8 (approved):
I still see the links "OpenAPI introduction" for example: https://docs-g60jtzv0z-ory.vercel.app/docs/sdk/new?language=go
docs-g60jtzv0z-ory.vercel.app[Ory Network SDK Reference | Ory](https://docs-g60jtzv0z-ory.vercel.app/docs/sdk/new?language=go)Index of REST operations from the generated OpenAPI docs, grouped by tag.[6:04 PM]How about replacing this text: "Use the links below to jump to a specific operation, or start from the OpenAPI introductionm" with this instead "Use the links below to jump to a specific operation or use the search bar to find an operation."

PR #2588, opened 2026-05-26
docs: rewrite OPL as practical guide

DavudSafarli:
This PR rewrites OPL into a practical how-to guide covering namespaces, relations, subject-set references, and permits, each section with examples


docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
+Every class must implement `Namespace`.
 
-### Comments
+## Relations
 
-1. Line comments start with the character sequence `//` and stop at the end of the line.
-2. General comments start with the character sequence `/*` and stop with the first subsequent character sequence `*/`.
-3. Documentation comments start with the character sequence `/**` and stop with the first subsequent character sequence `*/`.
+The `related` block defines who can be associated with an object. Each entry names a relation and declares what subject types it
+holds.
```
zepatrik: I think this can be confusing, as it is not necessarily clear what objects and subjects are here. I would either add something to namespaces section, or define object and subject here as the direction of relations.

Something along the lines:

```suggestion
The `related` block defines the relations between namespaces (classes). Each relation has a direction: `subject` is in `relation` of `object`. The relations defined in an OPL class (object) declare the allowed subjects. Generally, relations should represent the "real-world" relations between objects and subjects.
```

Then add something after the example saying which relations this defines and what the objects/subjects are.

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-Identifiers name program entities such as variables and types. An identifier is a sequence of one or more letters and digits. The
-first character in an identifier must be a letter.
+```ts
+import { Namespace } from "@ory/keto-namespace-types"
 
-```ebnf
-identifier = letter { letter | digit } .
-digit      = "0" … "9" .
-letter     = "A" … "Z" | "a" … "z" | "_" .
+class File implements Namespace {
```
zepatrik, line 38: class User missing

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-
-The top level of the configuration consists of a list of `class` declarations for each namespace. Each `class` consists of
-relation declarations and permission declarations.
-
-```ebnf
-Config          = [ ClassDecl ] .
-ClassDecl       = "class" identifier "implements" "Namespace" "{" ClassSpec "}" .
-ClassSpec       = [ RelationDecls ] | [ PermissionDefns] .
-```
+### Subject-set references
```
zepatrik, line 69: I propose to completely omit this section, as subject sets are not really useful currently. There is no specific reason to use them, right?

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-RelationType    = SubjectType | SubjectSetType .
-SubjectType     = TypeName .
-SubjectSetType  = "SubjectSet" "<" TypeName, string_lit ">" .
-TypeName        = identifier .
+```keto-natural
+members of Group:engineering is in viewers of File:readme
 ```
 
-Note that all relations are defined as array types `T[]` because there are naturally only many-to-many relations in Keto.
+A subject can now be either a `User` directly, or any member of the `Group:engineering`.
```
zepatrik: Not clear to me what this means.

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
 
-The following declares a type `Document` with three relations: `owners` and `viewers`, both of which have `users` as subjects.
-Additionally, the relation `parent` has type `Document`.
+## Permits
 
-```ts
-class User {}
+The `permits` block defines boolean functions the engine evaluates during a check.
```
zepatrik:
```suggestion
The `permits` block defines permissions. These are evaluated during a permission check request. Permissions are written as functions that return a boolean representing whether a permission is granted or not. While relations follow the real-world, permissions define application-specific rules based on relations.
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
 
-```ebnf
-PermissionDefns = "permits" "=" "{" Permission [ "," Permission ] "}" .
-Permission      = PermissionSign "=>" PermissionBody .
-PermissionSign  = PermissionName ":" "(" "ctx" [ ":" "Context" ] ")" [ ":" "boolean" ] .
-PermissionName  = identifier .
-```
-
-The `ctx` object is a fixed parameter that contains the `subject` for which the permission check should be conducted:
+`this.related.x.traverse(g => ...)` iterates over the objects in relation `x` and evaluates the inner expression for each one.
```
zepatrik:
```suggestion
`this.related.x.traverse(g => ...)` iterates over the subjects in relation `x` and evaluates the inner expression for each one.
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-  - `P` is a permission defined for all types referenced by `R`.
-- Given a `TransitiveCheck` as `this.related.R.transitive(x = x.related.S.includes(ctx.subject))`, we check that
-  - `R` is a relation defined for the current namespace and that
-  - `S` is a relation defined for all types referenced by `R`.
+A permission can call another permission defined on the same namespace:
 
-## Examples
+```ts
+edit:  (ctx: Context) => this.related.owners.includes(ctx.subject),
+admin: (ctx: Context) => this.permits.edit(ctx) && this.related.admins.includes(ctx.subject),
```
zepatrik: I think the example is not the most natural one
```suggestion
isAdmin: (ctx: Context) => this.related.admins.includes(ctx.subject),
edit:  (ctx: Context) => this.permits.isAdmin(ctx) ||this.related.owners.includes(ctx.subject),
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-    manager: User[]
-  }
-}
+class User implements Namespace {}
 
 class Group implements Namespace {
   related: {
-    members: (User | Group)[]
+    members: (User | SubjectSet<Group, "members">)[]
   }
```
zepatrik, line 194: While the subject set works, I don't think it is the correct way of doing this. Having the parents explicitly here makes it a lot clearer and allows the relation to be reused in different permissions if one wants to add them (think admins of groups, ...)

```suggestion
  related: {
    members: User[]
    parents: Group[]
  }
  permits = {
    isMember: (ctx: Context) => this.related.members.includes(ctx.subject) || this.related.parents.traverse((parent) => parent.permits.isMember(ctx))
```

zepatrik: I admit that it is not so nice in the other namespaces without the planned type assertions, but I think it is the better way of writing permissions.

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,183 @@
...
-    rename: (ctx: Context) => this.related.siblings.traverse((s) => s.permits.edit(ctx)),
   }
 }
 ```
+
+This schema models:
+
+- Direct access via `viewers` and `owners`
+- Group-based access via `SubjectSet<Group, "members">`
+- Inherited access from parent folders via `traverse`
```
zepatrik: Consider adding some of these details as comments, so it is more clear what you refer to. Consider adding more comments to the full example in general.

zepatrik (changes requested):

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: good point. Based on how we implement things, currently subjectsets are faster than traverse calls. In a place that traverse would cause 100 direct check calls, SubjectSet<> would make 1 call and get the answer.

But that's an implementation details, and technically, we **can** make optimizations to make them equally performant.

I think that's the only difference. Otherwise, Traverse is more future-proof as we can easily update/change the relationships form `.members.includes(.)` to `.admin.includes(.)`.

DavudSafarli: Okay, see this answer: https://github.com/orgs/authzed/discussions/2851#discussioncomment-15608216

This is also one difference, though i don't think anyone would care about it.

zepatrik: I see, but the use-case is very narrow.
IMO the only relevant use-case is where the relation is dynamic, so the user can choose whether they want to grant permissions to all group members, only group admins, ... That requires the relation to not be in the OPL though, or allow narrowing it down to multiple relations. The use-case is not supported really rn.

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: moving to separate User and Group relations, this cascades to other namespaces as well; now we can't do:
```
class Folder implements Namespace {
  related: {
    viewers: (User | SubjectSet<Group, "members">)[]
  }
```

instead we have to do:
```
userViewers: User[]
groupViewers: Group[]
```

and have a permit that does the OR logic. same goes for `File`, but more complex to read, bc it has `viewers` and `owners` relations, so it will become `viewerUsers`, `viewerGroups`, `ownerUsers`, `ownerGroups`.

We can remove the inheritance logic, and have simpler example to avoid this.

I guess this is "simplicity" that SubjectSets brings.

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: i improved the wording as:
```
This means a viewer can be either a `User` directly, or any subject in the `members` relation of a `Group`. You can then write a tuple that grants access to a whole group at once:
```

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: The final example makes use SubjectSets, and while changing it to avoid using SubjectSets, i had difficulty. I think there's still some value in mentioning this. Also, people can be migrating from a competitor to keto, and they have a working schema in their mind that uses SubjectSet.  I think it's worth mentioning this.

DavudSafarli:
@zepatrik i replied to your comments. There are few things to resolve about subjectsets and final example.

It's worth noting that reader ideally should read the other pages such as Namespaces/Objects/Subjects before jumping into OPL page. So, some understanding of the Namespace/Object/Subject should exist while reading this. This page is only for OPL reference. Not 1 page Keto guide.

Though, the other pages also need improvement, and overall sidebar needs improvements. As a reader, i alway preferred 1 long page docs compared to small separate pages. 
Also, if we have more "how to model rbac/gdrive/notion" guides, then there are more changes for users/llms to understand the OPL to build the right model.

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,218 @@
...
 ---
 
-Enforcing fine-grained permissions is a critical building block of mature technology solutions that protect privacy and identity
-in the information age. Several proprietary languages used to represent permission already exist, such as Rego or Casbin. Most
-permissions are defined by developers who are likely familiar with Web technologies like JavaScript or Typescript. There is a need
-for a developer-friendly configuration language for permissions that has a learning curve small enough so that most developers can
-understand and use it with minimal effort. To fulfill this need, we defined the permissions configuration language as a subset of
-the most common general-purpose programming language: JavaScript/TypeScript.
+Ory Permissions uses a relationship-based access control model (ReBAC): permissions are derived from the relationships between
+objects and subjects stored in the system. OPL is the TypeScript-based language you use to define those relationships and the
```
unatasha8:
```suggestion
objects and subjects stored in the system. Ory Permission Language (OPL) is the TypeScript-based language you use to define those relationships and the
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,218 @@
...
+permission rules that use them to control access.
 
-The Ory Permission Language is a syntactical subset of TypeScript. Along with type definitions for the syntax elements of the
-language (such as `Namespace` or `Context`), users can get context help from their IDE while writing the configuration.
+You write OPL to define things like "who can view a file", "whether a group member inherits access", or "whether owning a folder
```
unatasha8:
```suggestion
You use OPL to define a schema that includes permissions like "who can view a file", "whether a group member inherits access", or "whether owning a folder
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,218 @@
...
-2. General comments start with the character sequence `/*` and stop with the first subsequent character sequence `*/`.
-3. Documentation comments start with the character sequence `/**` and stop with the first subsequent character sequence `*/`.
+Here, `File:readme` is the object, `User:alice` and `User:bob` are the subjects, and `viewers` and `owners` are the relations.
 
-### Identifiers
+### Multiple subject types
 
-Identifiers name program entities such as variables and types. An identifier is a sequence of one or more letters and digits. The
-first character in an identifier must be a letter.
+Use a union when a relation can hold subjects of different types:
```
zepatrik:
```suggestion
Use a union type when a relation can hold subjects of different types:
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,218 @@
...
+### Subject-set references
 
-- `Context`
-- `Namespace`
-- `Namespace[]`
-- `boolean`
-- `string`
-- `SubjectSet<T, R>`
+Sometimes you want a relation to include not just individual subjects, but all members of another relation. `SubjectSet<T, R>`
+lets you do this — it refers to all subjects in relation `R` on namespace `T`.
```
zepatrik:
```suggestion
Sometimes you want a relation to include not just individual subjects, but all members of another relation. `SubjectSet<N, R>`
lets you do this — it refers to all subjects in relation `R` on namespace `N`.
```

docs/keto/reference/ory-permission-language.mdx
zepatrik: IMO that simplicity only comes from current limitations, not from an actually better approach. I'm a bit hesitant to prefer this (even if only for some time until we fix this), because people will use this now to write their OPL. Also, it might stick around for a lot longer when LLMs pick this up and recommend it, even when we later change this section. WDYT? Is it worth it to keep this?

zepatrik (commented):
I'd say we're very close here. Probably the language needs a brush-up, but that can also be done by the docs people just rephrasing what is currently here.

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: I updated it based on the text convo we had on slack, about the usefulness of subjectsets. it now says
```
For example, suppose a File can be shared with all members of a group, or just its admins:
```
for demostrating subjectsets

docs/keto/reference/ory-permission-language.mdx
DavudSafarli: done. this section is also removed

unatasha8 (commented):
See my comments from Slack

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
 sidebar_label: Ory Permission Language
 ---
 
-Enforcing fine-grained permissions is a critical building block of mature technology solutions that protect privacy and identity
-in the information age. Several proprietary languages used to represent permission already exist, such as Rego or Casbin. Most
-permissions are defined by developers who are likely familiar with Web technologies like JavaScript or Typescript. There is a need
-for a developer-friendly configuration language for permissions that has a learning curve small enough so that most developers can
-understand and use it with minimal effort. To fulfill this need, we defined the permissions configuration language as a subset of
-the most common general-purpose programming language: JavaScript/TypeScript.
+Ory Permissions uses a relationship-based access control model (ReBAC): permissions are derived from the relationships between
```
unatasha8:
```suggestion
Ory Keto (Permissions) uses a relationship-based access control model (ReBAC) in which permissions are derived from the relationships between
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-The Ory Permission Language is a syntactical subset of TypeScript. Along with type definitions for the syntax elements of the
-language (such as `Namespace` or `Context`), users can get context help from their IDE while writing the configuration.
+You write OPL to define things like "who can view a file", "whether a group member inherits access", or "whether owning a folder
+grants access to its contents". The schema you define is evaluated by the Ory Permissions engine at check time.
 
-## Notation
+## Namespaces
 
-The syntax is specified using the Extended Backus-Naur Form (EBNF):
+Each `class` in OPL defines a namespace — a type of object in your system, such as a file, folder, organization, or user.
```
unatasha8:
```suggestion
In OPL, each `class` defines a namespace, which represents a type of object in your system, such as a file, folder, organization, or user.
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-|   alternation
-()  grouping
-[]  option (0 or 1 times)
-{}  repetition (0 to n times)
-```
+## Relations
 
-Lowercase production names are used to identify lexical tokens. Non-terminals are in CamelCase. Lexical tokens are enclosed in
-double quotes `""` or single quotes `''`.
+In OPL, **object** refers to the thing being accessed (for example, a `File`), and **subject** refers to the entity requesting
```
unatasha8:
```suggestion
In OPL, `object` refers to the thing being accessed (for example, a `File`), and `subject` refers to the entity requesting
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
+class File implements Namespace {
+  related: {
+    viewers: User[]
+    owners: User[]
+  }
+}
+```
 
-## Lexical elements
+Relations are always arrays because an object can have many subjects. This declaration allows creating relationships like:
```
unatasha8:
```suggestion
Relations are always arrays because an object can have many subjects. For example, `File:readme` is the object, `User:alice` and `User:bob` are the subjects, and `viewers` and `owners` are the relations.


```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-### Comments
+```keto-natural
+User:alice is in viewers of File:readme
+User:bob is in owners of File:readme
+```
 
-1. Line comments start with the character sequence `//` and stop at the end of the line.
-2. General comments start with the character sequence `/*` and stop with the first subsequent character sequence `*/`.
-3. Documentation comments start with the character sequence `/**` and stop with the first subsequent character sequence `*/`.
+Here, `File:readme` is the object, `User:alice` and `User:bob` are the subjects, and `viewers` and `owners` are the relations.
```
unatasha8: Moved this line to above snippet. Delete here.

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-```ebnf
-identifier = letter { letter | digit } .
-digit      = "0" … "9" .
-letter     = "A" … "Z" | "a" … "z" | "_" .
+```ts
+viewers: (User | Group)[]
 ```
 
-### String literals
+This allows writing tuples with either a `User` or a `Group` as the subject:
```
unatasha8:
```suggestion
This allows writing relation tuples with either a `User` or a `Group` as the subject:
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-
-The top level of the configuration consists of a list of `class` declarations for each namespace. Each `class` consists of
-relation declarations and permission declarations.
-
-```ebnf
-Config          = [ ClassDecl ] .
-ClassDecl       = "class" identifier "implements" "Namespace" "{" ClassSpec "}" .
-ClassSpec       = [ RelationDecls ] | [ PermissionDefns] .
-```
+When writing tuples, you specify exactly which relation of the group to use as the subject:
```
unatasha8:
```suggestion
When writing relation tuples, you specify exactly which relation of the group to use as the subject:
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-```ebnf
-RelationDecls   = "related" "=" "{" { RelationName ":" ArrayType } "}" .
-RelationName    = identifier .
-ArrayType       = RelationType | ( "(" RelationType { "|" RelationType } ")" ) "[]" .
-RelationType    = SubjectType | SubjectSetType .
-SubjectType     = TypeName .
-SubjectSetType  = "SubjectSet" "<" TypeName, string_lit ">" .
-TypeName        = identifier .
-```
+The `permits` block defines permissions — functions that return a boolean, evaluated when a permission check is made. While
```
unatasha8:
```suggestion
The `permits` block defines permission functions, which return a boolean and are evaluated when a permission check is made. While
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
+relations model real-world associations, permissions define application-specific rules built on top of them.
 
-Note that all relations are defined as array types `T[]` because there are naturally only many-to-many relations in Keto.
-
-The following declares a type `Document` with three relations: `owners` and `viewers`, both of which have `users` as subjects.
-Additionally, the relation `parent` has type `Document`.
+Each permission function receives a `Context` object as its argument. `ctx.subject` refers to the entity whose access is being
+checked — the same subject used in relation tuples.
```
unatasha8:
```suggestion
checked, which is the same subject used in relation tuples.
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
 }
 ```
 
-### Permission definition
+To check membership within a permission function, OPL provides two methods: `includes` for direct membership, and `traverse` for
+inherited membership through another relation.
 
-Permissions are defined as functions within class declarations that take a parameter `ctx` of type `Context` and evaluate to a
-boolean `true` or `false`.
+### Direct membership: `includes`
```
unatasha8:
```suggestion
### Subject direct membership
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-The type annotations for `ctx` and the return value are optional.
+`this.related.<relation>.includes(ctx.subject)` checks whether the subject is directly in relation `<relation>`.
 
-```ebnf
-PermissionDefns = "permits" "=" "{" Permission [ "," Permission ] "}" .
-Permission      = PermissionSign "=>" PermissionBody .
-PermissionSign  = PermissionName ":" "(" "ctx" [ ":" "Context" ] ")" [ ":" "boolean" ] .
-PermissionName  = identifier .
-```
+### Inherited membership: `traverse`
```
unatasha8:
```suggestion
### Subject inherited membership
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-- a `IncludesCheck`, a check that something is in a set, e.g., `this.related.viewers.includes(ctx.subject)`:
-
-  ```ebnf
-  IncludesCheck = Var "." "related" "." RelationName "." "includes" "(" "ctx" "." "subject" ")" .
-  Var           = identifier .
-  ```
+`view` is granted if the subject is a member of any group in `viewerGroups`.
 
-- a `TranstitiveCheck`, a call to a permission on a relation, e.g., `this.related.parents.transitive(p => p.permits.view(ctx))`:
+### Boolean operators
```
unatasha8:
```suggestion
### Combine boolean operators
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-| Keto Config                                               | Zanzibar AST                                                                         |
-| --------------------------------------------------------- | ------------------------------------------------------------------------------------ |
-| `this.related.R.includes(ctx.subject)`                    | `computed_userset { relation: "R" } }`                                               |
-| `this.related.R.transitive(x = x.permits.P(ctx.subject))` | `tuple_to_userset { tupleset { relation: "R" } computed_userset { relation: "P" } }` |
+  restricted: (ctx: Context) => this.related.allowlist.includes(ctx.subject) && !this.related.blocklist.includes(ctx.subject),
+}
+```
 
-## Type checking
+### Calling another permission
```
unatasha8:
```suggestion
### Call another permission
```

docs/keto/reference/ory-permission-language.mdx
```diff
@@ -1,324 +1,247 @@
...
-- Given a `TransitiveCheck` as `this.related.R.transitive(x = x.related.S.includes(ctx.subject))`, we check that
-  - `R` is a relation defined for the current namespace and that
-  - `S` is a relation defined for all types referenced by `R`.
+```ts
+isAdmin: (ctx: Context) => this.related.admins.includes(ctx.subject),
+edit:  (ctx: Context) => this.permits.isAdmin(ctx) || this.related.owners.includes(ctx.subject),
+```
 
-## Examples
+## Complete example
```
unatasha8:
```suggestion
## Complete example for a permissions schema model
```

unatasha8 (changes requested):

unatasha8 (approved):
LGTM

zepatrik (approved):
Lets merge this, we can still improve it later. It is good for now.

PR #2603, opened 2026-06-04
docs: merge updated Ory Talos docs with path-based URLs

aeneasr:
Replaces the public Ory Talos documentation with the updated upstream content, switching to cleaner path-based page IDs across concepts, integrate, operate, quickstart, and reference (renaming the quickstarts to `open-source`/`docker-commercial`, adding `deploy/deployment-modes` and `security/admin-protection`, converting `architecture` and `edge-proxy` to MDX, and dropping `token-derivation-security`, `kubernetes`, and `separate-planes`). The API reference is regenerated from the new OpenAPI v3 spec (`reference/api.json`), adding the `ApiKeys` group and methods like `admin-revoke-issued-api-key`, `admin-revoke-imported-api-key`, `admin-batch-create-imported-api-keys`, and `get-jwks`. The OEL and OSS Talos sidebars are rewired to the new path-based IDs while preserving their existing placement and icon, and 38 redirects are added in `docusaurus.config.ts` from the old custom-ID URLs. `/talos` is kept as the Talos entrypoint and the Network sidebar is untouched; `npm run build` passes with no Talos-related broken links.

Documentation improvement; no issue reference required.

## Checklist

## Further comments

Documentation-only change. The `operate/benchmarks` page is intentionally left out of the OEL/OSS sidebars (matching upstream), but remains reachable at `/talos/operate/benchmarks` and its old URL redirects there.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Admin vs Public surfaces and recommended separate admin/public deployment
  * Expanded credential types, token format, security model, and secrets guidance
  * Reworked caching, rate-limiting, IP-restrictions, health, monitoring, and troubleshooting docs
  * Updated API reference (full OpenAPI), JWKS, and many CLI pages/examples
  * Removed/consolidated legacy deploy guides (Kubernetes, edge-proxy, separate-planes) and reorganized sidebars/redirects


docusaurus.config.ts
```diff
@@ -206,6 +206,190 @@ const config: Config = {
             from: "/product-selector",
             to: "/welcome",
           },
+          // Ory Talos: old custom-id URLs -> Rabat path-based URLs.
```
aeneasr, line 209: Critically verify if there are any URLs missing here or we have typos or URLs ended up here that we didn't need.

sidebars-oel.ts
```diff
@@ -475,46 +475,44 @@ const oelSidebar = [
             type: "category",
             label: "Quickstart",
             items: [
-              "talos/quickstart/index",
```
aeneasr, line 478: Programmatically check if we missed any pages, maybe with a linter.

docusaurus.config.ts
aeneasr: Verified programmatically by reconstructing every old custom-id URL from the `id:`/`slug:` frontmatter on `master` and cross-checking against the new path-based pages and this redirect list. Three issues found and fixed:

- **Removed 1 redirect we didn't need:** `/talos/quickstart/index` → that URL never existed (Docusaurus served the index page at the directory `/talos/quickstart`, which is already redirected the line above).
- **Added 2 missing redirects** for renamed/removed CLI reference pages:
  - `/talos/reference/cli/talos-keys-revoke` → `talos-keys-issued-revoke`
  - `/talos/reference/cli/talos-serve-check` → `talos-serve-public` (the data-plane verification server)

After the fix: 0 redirect targets that don't resolve, 0 phantom/typo `from:` entries, 0 old URLs left without a redirect. The 3 API renames already present check out.

sidebars-oel.ts
aeneasr: Ran a programmatic check: enumerated every Talos doc file and diffed against the ids referenced in `sidebars-oel.ts` (Talos is OEL-only — it's intentionally absent from the main `sidebars.ts`).

Found one regression: **`talos/operate/benchmarks`** was dropped from the sidebar. It was present on `master` as `talos/operate/talos-performance-benchmarks` (right after `configure`); the path-based rename renamed the file but not the sidebar entry. Re-added it in the same position. (The Operate index page links to it, so it was a real orphan.)

All other content pages are present. The remaining unreferenced files are the auto-generated `reference/api/*` and `reference/cli/*` pages, which are entry-point-only via `ory-talos-api` and `cli/talos` — same pattern as before this PR, so left as-is.

vinckr (dismissed):
- all files should be .mdx right? 
- since when do we use `description` in the frontmatter?

aeneasr:
Both answered from repo conventions:

## 1. No — not everything should be `.mdx`

The repo deliberately uses both: **720 `.md`** and **585 `.mdx`** files. Docusaurus runs both through the MDX loader, so the distinction is by content, not capability:

## 2. `description:` is not new — it's long-standing convention

https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter

PR #2616, opened 2026-06-09
feat: add tip to separate invite from recovery

zepatrik:


simi:
Adds a tip to the Invite users guide showing how to pass return_to on POST /admin/recovery/link, so admins can control where an invited user lands after setting credentials — and brand a first-time invitation differently from a normal password reset (e.g. `?return_to=…?source=invitation`).

`return_to` on this endpoint is supported but undocumented on the page where people implementing invitations actually look. It was added to the API spec in ory/kratos#3711 and requested in ory/kratos#3084, yet the invite-users guide never mentioned it. This closes that doc gap and notes the `allowed_return_urls` allow-list requirement.

vinckr:
the javascript tests fail because of missing auth, but they aren't relevant in this case. 

can just force merge it @aeneasr ?

vinckr (approved):

vinckr:
thanks @simi 🙌

PR #2639, opened 2026-06-22
docs: document at-most-once processing of OAuth2 logout requests

alnr:
## What changed

Updates the OpenID Connect logout documentation to reflect the faster,
database-free logout flow shipping in ory-corp/cloud#11260.

- **Logout logic diagram:** adds the reuse-detection branch. When the login
  session is already gone — because the `logout_verifier` was reused or the
  session was revoked by another logout, an API call, or expiry — Ory only
  redirects to the `post_logout_redirect_uri` instead of running the logout
  routines again.
- **New section "Logout requests are processed at most once":** explains the
  behavior in prose and clarifies that front- and back-channel logout callbacks
  fire at most once per session, so clients can treat them as idempotent.

## Why

In the new logout flow, logout challenges and verifiers are self-contained
tokens rather than database rows, so there is no longer a "handled" state for a
logout request. Fetching an already-handled logout request via
`GET /admin/oauth2/auth/requests/logout` no longer returns `410 Gone` (that
change flows into the published API reference automatically from the spec). The
user-facing guarantee — front/back-channel callbacks run at most once — is now
documented explicitly.

Companion to ory-corp/cloud#11260.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/oauth2-oidc/logout.mdx
+++ b/docs/oauth2-oidc/logout.mdx
@@ -126,8 +126,22 @@ RPI-->|has session cookie|LUI
 RPI-->|no session cookie|RPI2
 LUI-->|ask for user consent|LUI
 LUI-->|redirect with logout_verifier|RPI2[Process logout request]
-RPI2-->|execute front/backchannel logout, delete session cookie|RPI2
-RPI2-->|No post_logout_redirect_uri|End
-RPI2-->|Has post_logout_redirect_uri|End2[Redirect to post_logout_redirect_uri]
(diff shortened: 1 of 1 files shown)


vinckr (approved):

vinckr:
actually just noticed the mermaid diagram looks a little scuffed - is it supposed to look like this @alnr ?
I guess the loop at "ask for user consent" threw me a little. and some arrows are missing(?) - might be a mermaid thing....


Screenshot 2026-07-03 at 09 58 24

PR #2660, opened 2026-07-03
docs: note new identities default to active state

jhickmanit:
## What

Adds a short note to the **Create identities** page explaining that identities created via the admin API default to the \`active\` state (already visible as \`\"state\": \"active\"\` in the sample response, but previously unexplained), and how to create an \`inactive\` identity instead.

## Why

The sample response on this page shows \`\"state\": \"active\"\` with no explanation. This clarifies the default and points readers to the identity state page for activating/de-activating identities.

## Notes

- Verified the \`createIdentityBody\` API schema accepts a \`state\` field (enum \`active\`/\`inactive\`) in \`src/static/api.json\`, so the "set \`state\` to \`inactive\`" guidance is accurate.
- This section is still in the pre-migration \`docs/kratos/\` structure, so a plain \`/docs/...mdx\` file-path link is used (not \`SameDeploymentLink\`), consistent with neighboring pages. No migrated equivalent of this creation-flow page exists.

### DIFF
--- a/docs/kratos/manage-identities/20_create-users-identities.mdx
+++ b/docs/kratos/manage-identities/20_create-users-identities.mdx
@@ -74,6 +74,11 @@ The server response contains the created identity:
 }
 ```
 
+Newly created identities default to the `active` state, as shown by the `"state": "active"` field in the response. Active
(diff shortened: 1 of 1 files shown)


unatasha8 (approved):
LGTM

PR #2691, opened 2026-07-23
docs: correct iOS device-authentication guide

hperl:
Corrects three factual errors in the iOS device-authentication guide, found by comparing it against a reference iOS client implementation (ory-corp/cloud PR #12950).

- **iOS Simulator is usable.** The guide said the emulator cannot be used at all. In fact you can run device auth on the Simulator during development by enabling relaxed attestation, which accepts the software-backed attestation produced by the Ory Swift SDK's `OryClientSimulator` product. Relaxed attestation is refused once the project leaves the development environment.
- **App Attest keys cannot be biometric-gated through access control.** The guide said to create the signing key with `.biometryCurrentSet` in its access control. App Attest keys are system-managed and accept no access-control flags, so the client must run a `LocalAuthentication` check before every `generateAssertion` call. The server trusts the `platform` declaration on iOS because it cannot verify biometric gating remotely, which makes the client-side gate security-critical.
- **Keychain items survive app deletion.** The guide implied `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` causes uninstall to purge the stored blob. That attribute keeps the blob off encrypted backups and device migrations, not off a reinstall. A fresh install must detect the reinstall (for example via a `UserDefaults` install marker, which the system does clear) and drop any stale artifacts before use.

### DIFF
--- a/src/components/Shared/kratos/passwordless/deviceauthn/ios.mdx
+++ b/src/components/Shared/kratos/passwordless/deviceauthn/ios.mdx
@@ -2,10 +2,10 @@ A notable difference with Android is that Apple's app attestation APIs require a
 device.
 
 This means the iOS Simulator cannot produce real attestations. For development and testing you can still run on the Simulator by
-enabling <SameDeploymentLink to="kratos/passwordless/deviceauthn#relaxed-attestation-for-testing">relaxed attestation</SameDeploymentLink>,
-which accepts software-backed attestations. The Ory Swift SDK ships an `OryClientSimulator` product that generates such an
-attestation on the Simulator; its code compiles to nothing in device builds, so it can never ship in production. Relaxed
-attestation is refused once the project leaves the development environment.
+enabling <SameDeploymentLink to="kratos/passwordless/deviceauthn#relaxed-attestation-for-testing">relaxed
+attestation</SameDeploymentLink>, which accepts software-backed attestations. The Ory Swift SDK ships an `OryClientSimulator`
+product that generates such an attestation on the Simulator; its code compiles to nothing in device builds, so it can never ship
+in production. Relaxed attestation is refused once the project leaves the development environment.
 
 Since Device Binding only is supported on native devices (not in the browser), all corresponding API calls should be done using
 the endpoints for native apps, to avoid having to pass cookies around manually.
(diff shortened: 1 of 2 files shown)


src/components/Shared/kratos/passwordless/deviceauthn/ios.mdx
```diff
@@ -591,14 +599,20 @@ so its transport key is destroyed.
...
 
 - The attestation challenge is the **bare nonce**, not `SHA256(nonce ‖ t_pub)`. Pass the raw nonce bytes straight to `attestKey`
   as `clientDataHash`.
-- Create the signing key with `.biometryCurrentSet` in its access control, and enroll it with `"user_verification": "platform"`
-  and no `pin_protected` or `transport_public_key`.
+- Enroll with `"user_verification": "platform"` and no `pin_protected` or `transport_public_key`. There is no access-control flag
+  to set on the key; instead, gate signing with a `LocalAuthentication` check
+  (`LAContext.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics)`) that must succeed before every `generateAssertion` call,
+  at both enrollment and login. To approximate `.biometryCurrentSet` semantics — invalidating the key when the enrolled biometric
+  set changes — pin the context's `evaluatedPolicyDomainState` at enrollment and reject a login when it no longer matches.
```
pcaillaudm: I noticed `evaluatedPolicyDomainState` is marked deprecated in the [Apple docs](https://developer.apple.com/documentation/localauthentication/lacontext/evaluatedpolicydomainstate) - still fine to recommend here?

pcaillaudm (approved):
LGTM, just one non-blocking question

src/components/Shared/kratos/passwordless/deviceauthn/ios.mdx
```diff
@@ -1,7 +1,11 @@
 A notable difference with Android is that Apple's app attestation APIs require a network call to Apple's servers from a real
 device.
 
-This means that the emulator cannot be used.
+This means the iOS Simulator cannot produce real attestations. For development and testing you can still run on the Simulator by
+enabling <SameDeploymentLink to="kratos/passwordless/deviceauthn#relaxed-attestation-for-testing">relaxed
+attestation</SameDeploymentLink>, which accepts software-backed attestations. The Ory Swift SDK ships an `OryClientSimulator`
```
unatasha8:
```suggestion
attestation</SameDeploymentLink>, which accepts software-backed attestations. The Ory client SDK for Swift ships an `OryClientSimulator`
```

src/components/Shared/kratos/passwordless/deviceauthn/ios.mdx
```diff
@@ -424,8 +428,12 @@ func clientKeyId(fromUpdatedFlow flow: SettingsFlow) -> String? {
 }
 
 /// The local artifacts persisted after sealing. Store in the Keychain with
-/// kSecAttrAccessibleWhenUnlockedThisDeviceOnly — never in UserDefaults — so
-/// deleting the app purges them. None of these are secret on their own.
+/// kSecAttrAccessibleWhenUnlockedThisDeviceOnly — never in UserDefaults. The
+/// ThisDeviceOnly attribute keeps the blob off encrypted backups and device
+/// migrations, so it cannot follow the sealed secret onto another device.
+/// Keychain items survive app deletion, so detect a reinstall (for example with
```
unatasha8: It's a little hard to decipher whether 'so detect a reinstall' is a consequence or a direct action for the developer to do. Here is a rewrite of the complete comment. (I didn't do it inline as it's too hard to review the changes.)

///   Store this entire struct inside the Keychain using the attribute
///    'kSecAttrAccessibleWhenUnlockedThisDeviceOnly'. Never store it in UserDefaults.
///   The 'ThisDeviceOnly' flag prevents this data from being backed up 
///    or migrated. The artifacts will never follow a secret to another device.
///   Because the Keychain data survives deletion, you must write code to detect 
///  if the current launch is a reinstall. As the system completely wipes 
///  UserDefaults data when a user deletes the app (and creates a blank new one on reinstall), 
///  you can create a dummy marker/flag in UserDefaults to detect the reinstall. On boot up if 
/// a 'PinArtifacts' struct is in the Keychain, but your UserDefaults marker is missing, 
/// the user has reinstalled the app. You must delete these stale Keychain artifacts 
/// before allowing the app to run. (None of these struct fields are secret on their own).

unatasha8 (changes requested):

mm-ory (approved):

unatasha8 (approved):
LGTM

PR #2692, opened 2026-07-24
docs: update renamed deviceauthn fields references

pcaillaudm:
This PR updates the device authentication (`deviceauthn`) documentation to match the renamed settings-flow API fields.

Changes:

- Key registration: the `add` payload field is renamed to `deviceauthn_register`, and the SDK
  model `UpdateSettingsFlowWithDeviceAuthnMethodAdd` is now
  `UpdateSettingsFlowWithDeviceAuthnMethodRegister` (`method.add` → `method.deviceauthnRegister`).
- Key removal: the `delete` object with a nested `client_key_id` is replaced by a flat
  `deviceauthn_remove` string field, removing the need for the
  `UpdateSettingsFlowWithDeviceAuthnMethodDelete` wrapper in the SDKs.
- The Android and iOS registration examples now include the required `version: 1` field,
  aligning them with the payload reference in the overview page.

All JSON payloads, Kotlin/Swift SDK snippets, and prose references to the old field names
were updated accordingly.

### DIFF
--- a/src/components/Shared/kratos/passwordless/deviceauthn/android.mdx
+++ b/src/components/Shared/kratos/passwordless/deviceauthn/android.mdx
@@ -700,10 +700,10 @@ discipline when you wire these calls.
 
 1. **Enroll** (<SameDeploymentLink to="kratos/passwordless/deviceauthn#pin-enrollment">PIN enrollment</SameDeploymentLink>) —
    `decodeNonce` the flow's `deviceauthn_nonce` node, `createSealingKey`, create a `PinCeremony`, then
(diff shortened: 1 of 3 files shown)


unatasha8:
Isn't renaming fields in the API something that can break customer code? Have we added a note or given some heads up beyond updating the docs?

hperl (approved):

hperl:
> Isn't renaming fields in the API something that can break customer code? Have we added a note or given some heads up beyond updating the docs?

Your're right in principle, but device authentication has no users yet, so it was OK to do it now.

PR #2695, opened 2026-07-27
docs: add Console viewer role to roles and permissions

olagruchala:
Documents the read-only Viewer role in the Ory Console for both workspaces and projects.

  - Adds Viewer to the Workspace and Project role sections and to both permission matrices (Viewer / Developer / Owner).
  - Describes what a Viewer can read, and that resource-changing actions (save, create, delete, invite) are disabled with a tooltip.
  - Clarifies that Viewers cannot access runtime data managed by Ory services — identities, sessions, message delivery, OAuth2 clients, permission relationships, and issued & imported API keys.
  - Notes that project roles are additive: a member's effective project role is the higher of their workspace role and any role granted directly on the project. To give a workspace Viewer write access on one
  project, invite them to that project with the Developer role.

### DIFF
--- a/docs/console/roles-and-permissions.mdx
+++ b/docs/console/roles-and-permissions.mdx
@@ -8,7 +8,7 @@ The Ory Console uses role-based access control enforced through Ory Keto. Roles 
 
 ## Workspace roles
 
-A workspace has two roles: Owner and Developer.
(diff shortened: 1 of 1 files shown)


unatasha8 (approved):
LGTM

